PR #165, opened 2024-10-10
fix prometheus target and update dashboard to be more full.

aayushwhiz:
Changes Prometheus target in function calling demo so that it calls the correct endpoint since arch is no longer on same docker network. Updates dashboard to only include external stats and also display successful connection count and failed connection count.


adilhafeez:
@aayushwhiz can you pls post the screenshot of the dashboard here?

aayushwhiz:
> @aayushwhiz can you pls post the screenshot of the dashboard here?

Screenshot 2024-10-10 at 2 02 21 PM

adilhafeez (approved):

PR #183, opened 2024-10-13
Fix potential panic in create_http_context

junr03:
Signed-off-by: José Ulises Niño Rivera <junr03@users.noreply.github.com>

### DIFF
--- a/arch/src/filter_context.rs
+++ b/arch/src/filter_context.rs
@@ -282,7 +282,10 @@ impl RootContext for FilterContext {
         // No StreamContext can be created until the Embedding Store is fully initialized.
         let embedding_store = match self.mode {
             GatewayMode::Llm => None,
-            GatewayMode::Prompt => Some(Rc::clone(self.embeddings_store.as_ref().unwrap())),
+            GatewayMode::Prompt => match self.embeddings_store.as_ref() {
+                None => return None,
+                Some(store) => Some(Rc::clone(store)),
+            },
         };
         Some(Box::new(StreamContext::new(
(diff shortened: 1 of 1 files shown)


arch/src/filter_context.rs
```diff
@@ -282,7 +282,10 @@ impl RootContext for FilterContext {
         // No StreamContext can be created until the Embedding Store is fully initialized.
         let embedding_store = match self.mode {
             GatewayMode::Llm => None,
-            GatewayMode::Prompt => Some(Rc::clone(self.embeddings_store.as_ref().unwrap())),
```
junr03, line 285: this unwrap can panic if a stream is being created before the embeddings store has finished initializing

arch/src/filter_context.rs
```diff
@@ -282,7 +282,10 @@ impl RootContext for FilterContext {
...
+            GatewayMode::Prompt => match self.embeddings_store.as_ref() {
+                None => return None,
```
adilhafeez, line 286:
```suggestion
                None => None,
```

adilhafeez (approved):

arch/src/filter_context.rs
junr03: I did intend to `return None`. The whole point here is that a StreamContext should not be created if the Embedding Store is not present

adilhafeez: ah I see it now! make sense.

adilhafeez:
fixed in https://github.com/katanemo/arch/pull/194

PR #193, opened 2024-10-18
Pass tool call and app function response back in metadata

adilhafeez:
fixes https://github.com/katanemo/arch/issues/174


chatbot_ui/app/run.py
```diff
@@ -49,26 +52,35 @@ def predict(message, state):
...
     # this state must be passed back to the gateway in the next request
     response_json = json.loads(raw_response.text)
-    arch_state = None
     if response_json:
-        metadata = response_json.get("metadata", {})
-        if metadata:
-            arch_state = metadata.get(ARCH_STATE_HEADER, None)
-    if arch_state:
-        state["arch_state"] = arch_state
+        # load arch_state from metadata
```
salmanap, line 62: This is great - but I wonder if we can simplify this even further. Do we know if Gradio shows role=tool on the webui? I feel this code path will be the most trickest for developers to get right and will trip them up. And if we can avoid another js dependency for archgw in their code. Like envoy is fully transparent to the app, I wonder if we can get there too?

salmanap (changes requested):
Everything looks good - had one comment on state

salmanap (approved):

PR #195, opened 2024-10-18
Send back developer error correctly

adilhafeez:
This PR address the issue of gateway not sending application server errors back correctly. Before this change gateway would send 500 for any developer error. After this change developer should see correct error with error message in the body. Here is a sample 404 received by client that was sent by application server,

```
HTTP/1.1 404 Not Found
content-type: text/plain
date: Fri, 18 Oct 2024 03:50:37 GMT
server: envoy
transfer-encoding: chunked

63
upstream application error host=api_server, path=/weather, status=404, body={"detail":"some error"}
0
```

fixes https://github.com/katanemo/arch/issues/176


crates/prompt_gateway/src/prompt_stream_context.rs
```diff
@@ -164,6 +165,7 @@ impl PromptStreamContext {
         let embedding_response: CreateEmbeddingResponse = match serde_json::from_slice(&body) {
             Ok(embedding_response) => embedding_response,
             Err(e) => {
+                debug!("error deserializing embedding response: {}", e);
```
junr03, line 168: All of these debug logs should not be necessary. The error gets passed back to the user. Although if you do want debug logs printed to the console I would centralize that in send_server_error and simply print out the Display representation of the error struct.

junr03 (approved):
Bug fix lgtm. I would not pollute the code with decentralized debug logging.

crates/prompt_gateway/src/prompt_stream_context.rs
adilhafeez: Reason I put these logs out is so I can see the source of log. If logs are coming out from one function e.g. send_server_error then it would be quite difficult to find out the source of problem

junr03: The message should give you the source of the error.

PR #200, opened 2024-10-18
Update arch Dockerfile

junr03:
Fixes renamed path in #197.

@adilhafeez I think we should build some of the docker images on PR


adilhafeez (approved):

adilhafeez:
> @adilhafeez I think we should build some of the docker images on PR to prevent breakages like this

Yes, those changes are coming. We will build and publish docker imagaes from GH actions.

PR #202, opened 2024-10-18
Add support for streaming and fixes few issues (see description)

### LINKED ISSUE #172: Add support for compression/decompression
Right now filter doesn't understand compressed data. Fix here is to decompress data before it makes its way to filter and compress it on the way out depending on `accept-encoding` set by client. This is fairly simple task as envoy supports that out of the box.

### LINKED ISSUE #180: System prompt not sent to default LLM, when default promp_target not found
promp_target not found. Arch accepts an arch_config.yaml without a default target. But if the user sends a general QA prompt that doesn't match a specific prompt target, the system prompt is not being sent to the default LLM. This results in a behavior where the default LLM simply returns back its capabilities as compared to the capabilities being designed by the developer.

junr03:
In addition to adding support for streaming it adds following changes,

- handle compression of streams - fixes #172 
- adds e2e tests for llm and prompt gateway
- adds e2e tests as github action
- make prompt targets as option so we can run llm gateway without prompt targets
- adds support to resolve environment vars before generating final envoy yaml file #213
- update docker-compose.dev.yaml for local development
- add healthz endpoint in prompt gateway
- update chatbot ui to use streaming

also fixes #180


crates/common/src/common_types.rs
```diff
@@ -448,4 +505,173 @@ mod test {
             ParameterType::String
         );
     }
+
```
junr03, line 508: AMAZING use of testing!

salmanap (approved):
Reviewed over the shoulder. Overall an important and good change

PR #206, opened 2024-10-20
HR agent demo

salmanap:
Building an HR agent that can create workforce reports and trigger emails


demos/hr_agent/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
+services:
+  api_server:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      - SLACK_BOT_TOKEN=${SLACK_BOT_TOKEN:?error}
```
adilhafeez, line 7: This should be optional feature and we shouldn't fail. If token is empty then just print the rest api on console.

demos/hr_agent/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - "18083:80"
+    healthcheck:
+        test: ["CMD", "curl" ,"http://localhost:80/healthz"]
+        interval: 5s
+        retries: 20
+
+  chatbot_ui:
+    build:
+      context: ../../chatbot_ui
+      dockerfile: Dockerfile
```
adilhafeez, line 18: this is not required

demos/hr_agent/requirements.txt
```diff
@@ -0,0 +1,6 @@
+fastapi
+uvicorn
+pydantic
+slack-sdk
+typing
+pandas
```
adilhafeez, line 6: you're not using poetry?

demos/hr_agent/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+FROM base AS builder
+
+WORKDIR /src
+
+COPY requirements.txt /src/
+COPY workforce_data.json /src/
+RUN pip install --prefix=/runtime --force-reinstall -r requirements.txt
+
+COPY . /src
```
adilhafeez, line 11: ideally we should only copy files that are needed to be in image like copy only main.py and workforce_data.json

demos/hr_agent/arch_config.yaml
```diff
@@ -0,0 +1,63 @@
...
+listener:
+  address: 127.0.0.1
+  port: 8080 #If you configure port 443, you'll need to update the listener with tls_certificates
+  message_format: huggingface
+
+# Centralized way to manage LLMs, manage keys, retry logic, failover and limits in a central way
+llm_providers:
+  - name: OpenAI
+    provider: openai
+    access_key: OPENAI_API_KEY
```
adilhafeez, line 11: This isn't standard way we should use $OPEN_API_KEY for env replacement

adilhafeez: we should do sth similar to https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontendproxy/Dockerfile#L11

adilhafeez: filed issue https://github.com/katanemo/arch/issues/213

demos/hr_agent/main.py
```diff
@@ -0,0 +1,103 @@
...
+def send_slack_message(request: SlackRequest):
+    """
+    Endpoint that sends slack message
+    """
+    slack_message = request.slack_message
+
+    # Load the bot token from an environment variable or replace it directly
+    slack_token = os.getenv(
+        "SLACK_BOT_TOKEN"
+    )  # Replace with your token if needed: 'xoxb-your-token'
```
adilhafeez, line 68: if the token is empty should just print the call to console

demos/hr_agent/main.py
```diff
@@ -0,0 +1,103 @@
...
+
+class SlackRequest(BaseModel):
+    slack_message: str
+
+
+class WorkforceResponse(BaseModel):
+    region: str
+    staffing_type: str
+    headcount: int
+    satisfaction: float
```
adilhafeez, line 36: looks like this is not passed in from arch_config, as I didn't see it in parameters list for the prompt_target `workforce`

demos/hr_agent/main.py
```diff
@@ -0,0 +1,103 @@
...
+                    "role": "assistant",
+                    "content": "I am a helpful HR agent, and I can help you plan for workforce related questions",
+                },
+                "finish_reason": "completed",
+                "index": 0,
+            }
+        ],
+        "model": "hr_agent",
+        "usage": {"completion_tokens": 0},
+    }
```
adilhafeez, line 99: this should be simple, filed filed https://github.com/katanemo/arch/issues/214

demos/hr_agent/docker-compose.yaml
```diff
@@ -0,0 +1,25 @@
+services:
+  api_server:
+    build:
+      context: .
+      dockerfile: Dockerfile
```
adilhafeez, line 5: its not required if we're using default filename `Dockerfile`

adilhafeez (changes requested):
Inconsistent in naming used api_server and app_server at different places to refer to app_server

demos/hr_agent/requirements.txt
salmanap: simpler to use for demos

demos/hr_agent/main.py
salmanap: this is for response, not request

demos/hr_agent/Dockerfile
salmanap: Sure..

adilhafeez (approved):

PR #208, opened 2024-10-21
Cotran/hallucination

cotran2:
Fixing https://github.com/katanemo/arch/issues/205. Logic
- We reverse the messages list
- If the last assistant/system message is from Arch,
- We loop and concatenate all the user messages until it is not arch -> (gpt model)


adilhafeez (changes requested):
@cotran2  please add more tests to simulate intent change

adilhafeez (approved):

PR #209, opened 2024-10-21
Improve logging

adilhafeez:
Use consistent logging across prompt gateway codebase


crates/common/src/consts.rs
```diff
@@ -7,7 +7,7 @@ pub const SYSTEM_ROLE: &str = "system";
 pub const USER_ROLE: &str = "user";
 pub const TOOL_ROLE: &str = "tool";
 pub const ASSISTANT_ROLE: &str = "assistant";
-pub const GPT_35_TURBO: &str = "gpt-3.5-turbo";
+pub const DEFAULT_GPT_MODEL: &str = "gpt-3.5-turbo";
```
salmanap, line 10: Where is this used? Feels confusing.

crates/prompt_gateway/src/http_context.rs
```diff
@@ -316,14 +317,14 @@ impl HttpContext for StreamContext {
                             serde_json::Value::String(arch_state_str),
                         );
                         let data_serialized = serde_json::to_string(&data).unwrap();
-                        debug!("arch => user: {}", data_serialized);
+                        debug!("archgw <= developer: {}", data_serialized);
```
salmanap, line 320: should we update other parts of the debug logs to feel like

user => archgw
archgw => model_server (intent)
archgw <= application

Or is there an open standard that we can follow for our logs, so that developers can ingest them and augment them to have a clean view of the full application? I would prefer the latter so that we can demostrate that we are "open standards" based when it comes to logs and traces. We already do that for tracing via W3C Traceparent

salmanap: https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/application_logging#printing-logs-in-json-format

adilhafeez: We can do json logging which is easier for log agents to ingest that is a config change in envoy.yaml. That change can happen separately.

PR #217, opened 2024-10-23
fix model server stop process

cotran2:
PR to fix https://github.com/katanemo/arch/issues/207
Solution:

Increase timeout to 20s
Kill all children of the process
Kill all pids on port 51000

### DIFF
--- a/model_server/app/cli.py
+++ b/model_server/app/cli.py
@@ -15,9 +15,6 @@ logging.basicConfig(
 log = logging.getLogger("model_server.cli")
 log.setLevel(logging.INFO)
 
-# Path to the file where the server process ID will be stored
-PID_FILE = os.path.join(tempfile.gettempdir(), "model_server.pid")
-
 
 def run_server(port=51000):
     """Start, stop, or restart the Uvicorn server based on command-line arguments."""
@@ -38,11 +35,7 @@ def run_server(port=51000):
 
 
 def start_server(port=51000):
-    """Start the Uvicorn server and save the process ID."""
-    if os.path.exists(PID_FILE):
-        log.info("Server is already running. Use 'model_server restart' to restart it.")
-        sys.exit(1)
-
+    """Start the Uvicorn server"""
     log.info(
         "Starting model server - loading some awesomeness, this may take some time :)"
     )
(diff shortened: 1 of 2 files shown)


model_server/app/cli.py
```diff
@@ -105,12 +133,21 @@ def stop_server():
...
+        process.wait(timeout=20)  # Wait for up to 20 seconds for the process to exit
+
+        if process.is_running():
+            log.info(f"Process with PID {pid} is still running. Forcing shutdown.")
+            process.kill()  # Forcefully kill the process
 
         log.info(f"Model server with PID {pid} stopped.")
         os.remove(PID_FILE)
 
+        kill_process_on_port(port)
```
salmanap, line 149: curious, when will this be needed? If the PID file is not found? Then shouldn't we encapsulate that in the else block?

cotran2: PID file can be corrupted if you kept overwritting it. Happened to me and Adil so this function will be the last cleanse

cotran2: you are right too, im removing all pid now

model_server/app/cli.py
```diff
@@ -89,40 +79,63 @@ def wait_for_health_check(url, timeout=180):
...
+
+        # Step 2: Parse the process IDs from the output
+        process_ids = [line.split()[1] for line in result.stdout.splitlines()]
+
+        if not process_ids:
+            print(f"No process found listening on port {port}.")
+            return
+
+        # Step 3: Kill each process using its PID
+        for pid in process_ids:
```
salmanap, line 104: Would this work for Mac/Linux both? And I think we need to add test cases for this so that we are sure that this would work ever yime, all the time.

salmanap (changes requested):

adilhafeez (approved):
Make sure that `lsof` exists in linux. I think it does. Also what about windows? I know on linux you can do `netstat -p` to see process id associated with listening port. I do think windows also have similar netstat command to show pid.

cotran2:
We don't support windows, I check the windows solution, totally a different function.

model_server/app/cli.py
salmanap: Can we please add a test case for this. There is a test_cli.py now. There is a lot going on in this method and would be good to have some test coverage

cotran2: added test in latest pr

PR #218, opened 2024-10-23
Update chatbot UI and update hallucination check

nehcgs:


salmanap (approved):

cotran2:
Including new fix on the hallucination check:
1. Change system-role -> assistant-role (potential confusion for gpt if see system role in history)
2. add metadata for the hallucination chat completion request

Need review from @adilhafeez

adilhafeez:
@cotran2 why are we updating hallucination?

cotran2:
We observe a bug that if the role of the chat template for hallucination check is system role, it will cause gpt to be confused (correct here is assistant).

crates/prompt_gateway/src/hallucination.rs
```diff
@@ -18,9 +18,11 @@ pub fn extract_messages_for_hallucination(messages: &Vec<Message>) -> Vec<String
...
             if let Some(model) = message.model.as_ref() {
                 if !model.starts_with(ARCH_MODEL_PREFIX) {
-                    if message.role == ASSISTANT_ROLE {
-                        break;
-                    }
+                  if let Some(content) = &message.content {
+                      if !content.starts_with(HALLUCINATION_TEMPLATE) {
+                          break;
+                      }
+                  }
```
adilhafeez, line 25: is this needed

cotran2: Yes, this is for separating the hallucination template. When we collect the user history message, we want to go pass over the chat template if there is parameter gathering from ArchFC.

cotran2: Yes, this is for separating the hallucination template. When we collect the user history message, we want to go pass over the chat template if there is parameter gathering from ArchFC.

adilhafeez: can you add test case to simulate this behavior

cotran2: the current test case reflects this.

PR #227, opened 2024-10-29
Improve Gradio UI and fix arch_state bug

adilhafeez:
- fix arch_state bug that was causing hallucination to fail this was caused by https://github.com/katanemo/arch/issues/171
- don't send tool response and tool call to upstream LLMs. This was unintentionally introduced by https://github.com/katanemo/arch/pull/202
- improve gradio ui code to better handle stream conversations

image


salmanap (approved):
ship it

PR #229, opened 2024-10-29
add support for jaeger tracing

adilhafeez:
- move prompt and llm gateway wasm filter into single listener
- create trace_arch_internal param so we can expose arch internal spans for local development
- when making outbound calls from filter attach traceparent in the header
- create a new internal listener to generate traceparent so spans from filter can be grouped properly
- create signoz and jaeger shared service
- rename function_calling to weather_forecast

trace ui for a sample request,

image


arch/arch_config_schema.yaml
```diff
@@ -149,6 +149,8 @@ properties:
     properties:
       random_sampling:
         type: integer
+      trace_arch:
```
salmanap, line 152: trace_arch_internal please

demos/shared/insurance_claims_service/Dockerfile
```diff
@@ -0,0 +1,41 @@
+# took inspiration from https://medium.com/@albertazzir/blazing-fast-python-docker-builds-with-poetry-a78a66f5aed0
```
salmanap, line 1: Please remove this altogether. No need to keep this insurance_claims_service. We have a separate demo for that

demos/shared/insurance_claims_service/main.py
```diff
@@ -0,0 +1,83 @@
+import json
```
salmanap, line 1: Please remove this altogether. See comment above

demos/shared/insurance_claims_service/pyproject.toml
```diff
@@ -0,0 +1,23 @@
+[tool.poetry]
```
salmanap, line 1: Remove please. See above

salmanap (commented):
I

demos/shared/insurance_claims_service/main.py
adilhafeez: removed

demos/shared/insurance_claims_service/Dockerfile
adilhafeez: removed

demos/shared/insurance_claims_service/pyproject.toml
adilhafeez: removed

arch/arch_config_schema.yaml
adilhafeez: updated

model_server/app/main.py
```diff
@@ -15,12 +16,42 @@
 )
 from unittest.mock import patch
 
+from opentelemetry import trace
```
salmanap, line 19: Please remove tracing from model_server internal calls. Not necessary

adilhafeez: its not used - see this line https://github.com/katanemo/arch/pull/229/files#diff-a883497ecfcfe46bc8a13b33b682d89dc175ff777398e7f8fce68cfc990087f1R46

salmanap (approved):

PR #230, opened 2024-10-29
Use large github action machine to run e2e tests

adilhafeez:
- use large github action runner so that we don't run out of space
- use healthz check to wait for arch gateway before giving control back to user
- run e2e tests using archgw command
- fix code bug in arch gateway to remove tool call and app response when there is no match for prompt target
- add $ sign env vars
- add retry to pytests

image


salmanap (commented):
Looks good to me - lets add a description for the release please. And once we publish this to main, we should update all tools/images.

salmanap (approved):

PR #233, opened 2024-10-30
Demo fixes

dougalb:
Small fixes for demo to get them working.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 arch/qdrant_data/
 /venv/
+/.venv/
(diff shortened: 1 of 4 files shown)


adilhafeez (approved):
Thanks a lot for putting up this PR Dougal, +1'ed

adilhafeez:
Thanks Dougal for your PR. I've improved it by using "$" as key to resolve env vars. In the absence of $ sign we will treat the value as it is. https://github.com/katanemo/arch/pull/230

PR #236, opened 2024-10-31
add prefill and test

cotran2:
Pre fill mechanism to resolve issue https://github.com/katanemo/arch/issues/237
1. Check if first non empty token is "<tool_call>"
- If yes, continue to get the full response
- If no, prefill the assistant with filler "Sure!" to engage parameter gathering.
-

2. Add unit test for this mechanism and function calling utils


model_server/app/function_calling/model_utils.py
nehcgs, line 1: 1. Add `"continue_final_message": True` and `"add_generation_prompt": False` to `extra_body` in Line 120
2. Merge Line 89-98 with Line 130 - 140
3. Add `resp.close()` in `except Exception as e`

model_server/app/tests/test_function_calling.py
nehcgs, line 1: 1. Add one more test for prompt prefilling, checking whether the model response starts with the prefix chosen from the list.

cotran2: It is already in (last line)

model_server/app/function_calling/model_utils.py
cotran2: 89-98 is stream = True
130-140 is stream = False
Are we always gonna go stream = True?

cotran2: add resp.close() in expection will fail because the request doesn't come thru and resp object is not created yet.

nehcgs: For streaming, two options here:
1. Set `stream=True`:  then concatenate tokens and use `tokenizer` to clean up tokenization space. We will have to do this eventually to support parallel & multiple function calling.
2. Set `stream=prefill_enabled`, which will keep the current method, but simply the code by removing repeated code.

e2e_tests/test_prompt_gateway.py
```diff
@@ -260,3 +260,40 @@ def test_prompt_gateway_default_target(stream):
             response_json.get("choices")[0]["message"]["content"]
             == "I can help you with weather forecast or insurance claim details"
         )
+
+
+@pytest.mark.parametrize("prefill_enabled", [True, False])
```
adilhafeez, line 265: you may also want to add stream [true, false] to make sure this feature works in stream and non-stream case

e2e_tests/test_prompt_gateway.py
```diff
@@ -260,3 +260,40 @@ def test_prompt_gateway_default_target(stream):
...
+def test_prompt_gateway_arch_prefill(prefill_enabled):
+    body = {
+        "messages": [
+            {
+                "role": "user",
+                "content": "how is the weather",
+            }
+        ],
+        "prefill_enabled": prefill_enabled,
```
adilhafeez, line 274: this isn't right place as it would break the contract for openai. We should use overrides to configure prefill here
https://github.com/katanemo/arch/blob/main/demos/function_calling/arch_config.yaml#L14-L16

e2e_tests/common.py
```diff
@@ -30,7 +30,8 @@ def get_arch_messages(response_json):
     arch_messages = []
     if response_json and "metadata" in response_json:
         # load arch_state from metadata
-        arch_state_str = response_json.get("metadata", {}).get(ARCH_STATE_HEADER, "{}")
+        arch_state_str = response_json.get("metadata") or {}
```
adilhafeez, line 33: this would do the same 

`response_json.get("metadata", {})`

adilhafeez (changes requested):

e2e_tests/test_prompt_gateway.py
cotran2: Prefill enabled will make stream = True

adilhafeez: I mean steam for response from upstream. I do think it would have no effect on it as when we engage with parameter gathering we don't engage with upstream llm.

e2e_tests/test_prompt_gateway.py
cotran2: This only for the prompt gateway

e2e_tests/common.py
cotran2: I got a weird bug when response_json['metadata'] = Null

e2e_tests/test_prompt_gateway.py
cotran2: not sure how to add this tbh, prefill_enable dictate streaming

model_server/app/function_calling/model_utils.py
```diff
@@ -67,30 +81,75 @@ async def chat_completion(req: ChatMessage, res: Response):
...
+            logger.info("Tool call is not found! Engage pre filling")
+            prefill_content = random.choice(const.prefill_list)
+            messages.append({"role": "assistant", "content": prefill_content})
+
+            # Send a new completion request with the updated messages
+            extra_body = {
+                **const.arch_function_generation_params,
+                "continue_final_message": True,
+                "add_generation_prompt": False,
+            }
```
adilhafeez, line 119: pls add a comment here what does this extra_body do?

model_server/app/function_calling/model_utils.py
```diff
@@ -67,30 +81,75 @@ async def chat_completion(req: ChatMessage, res: Response):
...
+        first_token_content = ""
+        for token in resp:
+            first_token_content = token.choices[
+                0
+            ].delta.content.strip()  # Clean up the content
+            if first_token_content:  # Break if it's non-empty
+                break
+
+        # Check if the first token requires tool call handling
+        if first_token_content != "<tool_call>":
```
adilhafeez, line 107: create a const for `<tool_call>` and compare with it

model_server/app/commons/constants.py
```diff
@@ -8,6 +8,8 @@
 logger = utils.get_model_server_logger()
 
 arch_function_hanlder = ArchFunctionHandler()
+prefill_list = ["May", "Could", "Sure", "Definitely", "Certainly", "Of course", "Can"]
+prefill_enabled = True
```
adilhafeez, line 12: pls make consts in all caps (with underscore separating words)

adilhafeez (approved):
looks good - but pls address comments before merging

PR #245, opened 2024-11-04
retry embeddings fetch

adilhafeez:
Updated arch gateway code to not panic if embedding server is not up yet. Rather it will keep on trying.

This code change fixes https://github.com/katanemo/arch/issues/247 and https://github.com/katanemo/arch/issues/243

### DIFF
--- a/crates/prompt_gateway/src/filter_context.rs
+++ b/crates/prompt_gateway/src/filter_context.rs
@@ -12,7 +12,7 @@ use common::http::Client;
 use common::stats::Gauge;
 use common::stats::IncrementingMetric;
 use http::StatusCode;
-use log::{debug, info, warn};
+use log::{debug, info, trace, warn};
 use proxy_wasm::traits::*;
 use proxy_wasm::types::*;
 use std::cell::RefCell;
@@ -215,7 +215,7 @@ impl Context for FilterContext {
         body_size: usize,
         _num_trailers: usize,
     ) {
-        debug!(
+        trace!(
             "filter_context: on_http_call_response called with token_id: {:?}",
             token_id
         );
@@ -283,10 +283,7 @@ impl RootContext for FilterContext {
             context_id
         );
 
-        let embedding_store = match self.embeddings_store.as_ref() {
-            None => None,
-            Some(store) => Some(Rc::clone(store)),
-        };
+        let embedding_store = self.embeddings_store.as_ref().map(Rc::clone);
         Some(Box::new(StreamContext::new(
             context_id,
             Rc::clone(&self.metrics),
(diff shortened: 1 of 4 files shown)


crates/prompt_gateway/tests/integration.rs
```diff
@@ -332,7 +333,7 @@ llm_providers:
 
 overrides:
   # confidence threshold for prompt target intent matching
-  prompt_target_intent_matching_threshold: 0.6
+  prompt_target_intent_matching_threshold: 0.0
```
cotran2, line 336: why we changing this?

crates/prompt_gateway/tests/integration.rs
```diff
@@ -262,7 +263,7 @@ fn setup_filter(module: &mut Tester, config: &str) -> i32 {
         )
         .returning(Some(101))
         .expect_metric_increment("active_http_calls", 1)
-        .expect_set_tick_period_millis(Some(0))
+        .expect_set_tick_period_millis(Some(5000))
```
cotran2, line 266: why 5s?

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -234,7 +245,7 @@ impl StreamContext {
                             "description embeddings not found for prompt target name: {}",
                             prompt_name
                         );
-                        return (prompt_name.clone(), f64::NAN);
+                        return (prompt_name.clone(), 0.0);
```
cotran2, line 248: why return 0.0 instead of nan?

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -223,7 +234,7 @@ impl StreamContext {
                             "embeddings not found for prompt target name: {}",
                             prompt_name
                         );
-                        return (prompt_name.clone(), f64::NAN);
+                        return (prompt_name.clone(), 0.0);
```
cotran2, line 237: same

crates/prompt_gateway/src/http_context.rs
```diff
@@ -35,10 +35,10 @@ impl HttpContext for StreamContext {
 
         let request_path = self.get_http_request_header(":path").unwrap_or_default();
         if request_path == HEALTHZ_PATH {
-            if self.embeddings_store.is_none() {
-                self.send_http_response(503, vec![], None);
-            } else {
+            if self.is_embedding_store_initialized() {
```
cotran2, line 38: why do we need this

crates/prompt_gateway/src/filter_context.rs
```diff
@@ -66,22 +68,26 @@ impl FilterContext {
             prompt_guards: Rc::new(PromptGuards::default()),
             embeddings_store: Some(Rc::new(HashMap::new())),
             temp_embeddings_store: HashMap::new(),
+            active_embedding_calls_count: 0,
```
cotran2, line 71: i think this is very hard to track

cotran2 (changes requested):

crates/prompt_gateway/tests/integration.rs
adilhafeez: this is for testing, I changed nan  -> 0.0 when embeddings is not found. So had to lower the threshold so we go to weather endpoint

crates/prompt_gateway/tests/integration.rs
adilhafeez: this is wait before next retry

crates/prompt_gateway/src/stream_context.rs
adilhafeez: 0 tells there is no similarity, this only happens when embeddings are not found

crates/prompt_gateway/src/http_context.rs
adilhafeez: this is used in `archgw up` and in e2e tests to make sure we are healthy before users/developers start sending requests to us. See health [check here](https://github.com/katanemo/arch/blob/main/arch/docker-compose.yaml#L17-L21).

cotran2 (approved):

salmanap (approved):

Issue #246, opened 2024-11-04
[llm-gateway] add tft and tot metrics and add them to grafana dashboard


adilhafeez:
Add TFT and total request time in llm-gateway.

For TFT the timer should be started when the [request is received](https://github.com/katanemo/arch/blob/main/crates/llm_gateway/src/stream_context.rs#L141-L161) and timer should end then [first token is received](https://github.com/katanemo/arch/blob/main/crates/llm_gateway/src/stream_context.rs#L318) from upstream LLM.

The unit for metric should be milliseconds and should be stored in [WasmMetrics struct](https://github.com/katanemo/arch/blob/main/crates/llm_gateway/src/filter_context.rs#L16-L19).

salmanap:
We need the following metrics instrumented

1. Time to first token (TTFT): This is how quickly users start seeing the model’s output after entering their query. Low waiting times for a response are essential in real-time interactions, but less important in offline workloads. This metric is driven by the time required to process the prompt and then generate the first output token.
2. Time per output token (TPOT): Time to generate an output token for each user that is querying the system. 
3. Latency = TTFT + (TPOT) * (the number of tokens to be generated)
4. Output sequence length (OSL) - Total tokens generated
4. Input sequence length (ISL) - Total input tokens 

We need the following metrics histograms in 5 mins, 15 mins, 30 mins, 1 hour, 24 hours, 7 day granularity. 

1. Average (p50)
2. P90 (90th percentile)
3. p95 (95th percentile)
4. p99 (99th percentile)

PR #250, opened 2024-11-06
Update arch stats

aayushwhiz:
solves #246 by changing stats to implement debug for histogram, update filter_context to open ttft to stats endpoint and update stream_context to get time between both of those.


crates/llm_gateway/src/filter_context.rs
```diff
@@ -16,13 +17,23 @@ use std::rc::Rc;
 pub struct WasmMetrics {
     pub active_http_calls: Gauge,
     pub ratelimited_rq: Counter,
+    pub time_to_first_token: Histogram,
+    pub time_per_output_token: Histogram,
+    pub latency: Histogram,
```
adilhafeez, line 22: pls change latench => request_latency

crates/llm_gateway/src/stream_context.rs
```diff
@@ -120,16 +133,39 @@ impl StreamContext {
...
+        // Check if rate limiting needs to be applied.
         if let Some(selector) = self.ratelimit_selector.take() {
-            // Tokenize and Ratelimit.
-            if let Ok(token_count) = tokenizer::token_count(model, json_string) {
-                ratelimit::ratelimits(None).read().unwrap().check_limit(
-                    model.to_owned(),
-                    selector,
-                    NonZero::new(token_count as u32).unwrap(),
-                )?;
+            log::debug!("Rate limiting applied for model: {}", model);
```
adilhafeez, line 147:
```suggestion
            log::debug!("Applying rate limits for model: {}", model);
```

crates/llm_gateway/src/stream_context.rs
```diff
@@ -120,16 +133,39 @@ impl StreamContext {
...
+                        "Rate limit check failed for model: {} with error: {:?}",
+                        model,
+                        e
+                    );
+                    return Err(e);
+                }
             }
+        } else {
+            log::debug!("No rate limit applied for model: {}", model);
         }
```
adilhafeez, line 167: these debug statements can go away, the previous code seemed a bit cleaner

crates/llm_gateway/src/stream_context.rs
```diff
@@ -158,6 +194,20 @@ impl HttpContext for StreamContext {
...
+            Ok(current_time) => {
+                self.start_time = Some(current_time);
+                self.ttft_recorded = false;
+                self.ttft_duration = None;
+            }
+            Err(e) => {
+                warn!("Failed to get current time: {:?}", e);
+                self.start_time = None;
+            }
+        }
```
adilhafeez, line 209: It is unlikely that get_current_time would fail - but it failed we should fail the request

crates/llm_gateway/src/stream_context.rs
```diff
@@ -256,6 +306,32 @@ impl HttpContext for StreamContext {
 
         let body = if self.streaming_response {
             if end_of_stream && body_size == 0 {
+                // All streaming responses end with bytes=0 and end_stream=true
```
adilhafeez, line 309: what if its non-streaming response, would you still track total duration?

crates/llm_gateway/src/stream_context.rs
```diff
@@ -256,6 +306,32 @@ impl HttpContext for StreamContext {
...
+                                // Record the latency to the latency histogram
+                                self.metrics.latency.record(duration_ms as u64);
+                            }
+                            Err(e) => {
+                                warn!("SystemTime error: {:?}", e);
+                            }
+                        },
+                        Err(e) => {
+                            warn!("Failed to get current time: {:?}", e);
+                        }
```
adilhafeez, line 327: these are not necessary, fail the request if get_current_time fail

crates/llm_gateway/src/stream_context.rs
```diff
@@ -362,6 +507,31 @@ impl HttpContext for StreamContext {
...
+            //             },
+            //             Err(e) => {
+            //                 warn!("Failed to get current time: {:?}", e);
+            //             }
+            //         }
+            //         self.ttft_recorded = true;
+            //     } else {
+            //         warn!("Start time was not recorded");
+            //     }
+            // }
```
adilhafeez, line 534: remove dead code

crates/llm_gateway/src/stream_context.rs
```diff
@@ -344,6 +420,75 @@ impl HttpContext for StreamContext {
                     }
                 };
             self.response_tokens += token_count;
+
+            // Compute TTFT if not already recorded
+            if !self.ttft_recorded {
```
adilhafeez, line 425: You don't need separate `ttft_recorded`. You could use optional `ttft_duration` for the same purpose

crates/llm_gateway/src/stream_context.rs
```diff
@@ -344,6 +420,75 @@ impl HttpContext for StreamContext {
...
+                        }
+                    }
+                    self.ttft_recorded = true;
+                } else {
+                    warn!("Start time was not recorded");
+                }
+            }
+            // Check if first token was not processed yet, and if there are tokens in the response.
+            // If so, set the last_token_time to now and set first_token_processed to true
+            if !self.first_token_processed && token_count > 0 {
```
adilhafeez, line 450: don't need to maintain separate var for `self.first_token_processed`, could use optional last_token_time to achieve similar behavior

crates/llm_gateway/src/stream_context.rs
```diff
@@ -344,6 +420,75 @@ impl HttpContext for StreamContext {
...
+                                    // Convert the duration to milliseconds
+                                    let duration_ms = duration.as_millis();
+                                    debug!(
+                                        "Time for Current Output Token: {} milliseconds",
+                                        duration_ms as u64 / token_count as u64
+                                    );
+                                    // Record TPOT metric for historgram
+                                    self.metrics
+                                        .time_per_output_token
+                                        .record((duration_ms as u64) / (token_count as u64));
```
adilhafeez, line 477: This is compute intensive, and not sure there is lot of value in this. Can you please take it out for now. We need to measure its impact before we add it.

adilhafeez (approved):

crates/llm_gateway/src/stream_context.rs
adilhafeez: updated code

PR #255, opened 2024-11-08
update rust toolchain to 1.82

adilhafeez:
wasm-wasi32 target is deprated and will be removed in rust 1.84. Updating our build system to use wasm-wasi32p1 which is the recommendation. See [here for detail](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html).

Also update envoy to 1.32 and python to 3.12.

### DIFF
--- a/arch/Dockerfile
+++ b/arch/Dockerfile
@@ -1,22 +1,22 @@
 # build filter using rust toolchain
-FROM rust:1.80.0 as builder
-RUN rustup -v target add wasm32-wasi
+FROM rust:1.82.0 as builder
+RUN rustup -v target add wasm32-wasip1
 WORKDIR /arch
 COPY crates .
 
-RUN cd prompt_gateway && cargo build --release --target wasm32-wasi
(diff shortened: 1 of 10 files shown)


salmanap (approved):

arch/Dockerfile
```diff
@@ -1,22 +1,22 @@
...
+RUN cd prompt_gateway && cargo build --release --target wasm32-wasip1
+RUN cd llm_gateway && cargo build --release --target wasm32-wasip1
 
 # copy built filter into envoy image
-FROM envoyproxy/envoy:v1.31-latest as envoy
+FROM envoyproxy/envoy:v1.32-latest as envoy
 
 #Build config generator, so that we have a single build image for both Rust and Python
-FROM python:3-slim as arch
+FROM python:3.13-slim as arch
```
salmanap, line 14: Do we want to go to python 3.13. Isn't 3.12 the more stable version?

adilhafeez: using python:3.12 all across now

PR #262, opened 2024-11-12
Add service to stream custom otel traces to otel-collector

adilhafeez:
After this change we can generate custom spans showing total time taken by upstream llm

image


demos/shared/trace_streamer/stream_traces.py
```diff
@@ -0,0 +1,34 @@
...
+            data=line,
+            headers={"Content-Type": "application/json"},
+        )
+        print(f"Sent trace to otel-tracing: {response.status_code}")
+    except Exception as e:
+        print(f"Failed to send trace to otel-tracing: {e}")
+
+
+with open(envoy_log_path, "r") as f:
+    while True:
```
salmanap, line 27: I think best to check if the log file exists to begin with, if not that means tracing was off - then no need to run this for loop.

crates/prompt_gateway/src/http_context.rs
```diff
@@ -238,18 +243,69 @@ impl HttpContext for StreamContext {
...
+                    parent_span_id: Some(parent_span_id.to_string()),
+                    span_id: format!("{}", get_random_span_id()),
+                    name: "total_time".to_string(),
+                    start_time_unix_nano: format!("{}", self.start_upstream_llm_request_time),
+                    end_time_unix_nano: format!("{}", since_the_epoch_ns),
+                    kind: 1,
+                    attributes: vec![],
+                });
+
+                trace_data.add_span(Span {
```
salmanap, line 291: I think this should be an even in the total_time span. We shouldn't represent TTFT as a separate span because there is technically no new service work. https://opentelemetry.io/docs/concepts/signals/traces/#span-events

demos/shared/trace_streamer/stream_traces.py
adilhafeez: it must exist - if it doesnt we should fail/crash

I also make sure to create this file in Dockerfile see here -https://github.com/katanemo/arch/pull/262/files#diff-8e3a6a6de1bde2a67bc1791e073f4724332070eac7b751f83122f8929eb87f1fR9

salmanap (approved):

PR #267, opened 2024-11-14
move custom tracer to llm filter

adilhafeez:
- move trace_streamer to arch Dockerfile so we don't have to run another separate service
- apply trace hack to llm listener so we can get traceparent
- move ttft and other tracing changes to llm gateway
- add readme.md for llm_routing
- update gradio to add debug log and ability to override models

Sample trace,
image

Also updated Gradio UI a bit - 
image


arch/stream_traces.py
```diff
@@ -29,14 +30,10 @@ def process_log_line(line):
...
-        line = f.readline()
-        if not line:
-            time.sleep(1)
-            continue
-        tokens = line.split("prompt_gateway: upstream_llm trace details: ")
-        if len(tokens) > 1:
-            process_log_line(tokens[1])
+for line in sys.stdin:
+    if not line:
+        time.sleep(1)
```
cotran2, line 35: why we need sleep, this will make application very slow

crates/llm_gateway/src/stream_context.rs
```diff
@@ -53,6 +57,9 @@ impl StreamContext {
             request_id: None,
             start_time: None,
             ttft_duration: None,
+            traceparent: None,
+            ttft_time: None,
```
cotran2, line 61: probably not now but are we tracking other latency metrics? e2e for example, we should structure the parameters as a class, not just a parameter

demos/llm_routing/README.md
```diff
@@ -0,0 +1,19 @@
+# LLM Routing
+This demo shows how you can arch gateway to manage keys and route to appropricate LLM.
```
cotran2, line 2: how are we routing right now?

cotran2 (changes requested):

arch/stream_traces.py
adilhafeez: that is only when line is empty otherwise it will block forever

crates/llm_gateway/src/stream_context.rs
adilhafeez: this tracks time to first token. This whole code block was moved from prompt filter

demos/llm_routing/README.md
adilhafeez: client can configure upstream llm based on `x-arch-llm-provider-hint` header

adilhafeez: see code change in chatbot-ui

cotran2 (approved):

PR #269, opened 2024-11-14
Add in tpot

aayushwhiz:
Adds in time per output token as well as tokens per second metric.


adilhafeez (approved):
pls update dashboard to include this metric and then you can merge it

PR #278, opened 2024-11-18
fix start time of llm filter

adilhafeez:


salmanap (approved):
Do we need to add tests for this?

adilhafeez:
> Do we need to add tests for this?

yes we do, it should be captured in e2e tests

PR #281, opened 2024-11-18
hallucination with log probs

cotran2:


model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+    return entropy.item(), varentropy.item()
+
+
+def check_parameter_property(api_description, parameter_name, property_name):
+    """
+    Check if a parameter in an API description has a specific property.
+
+    Args:
+        api_description (dict): The API description in JSON format.
```
adilhafeez, line 56: api_description should be func property

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+    token_probs = torch.exp(log_probs)
+    entropy = -torch.sum(log_probs * token_probs, dim=-1) / math.log(2, math.e)
+    varentropy = torch.sum(
+        token_probs * (log_probs / math.log(2, math.e)) + entropy.unsqueeze(-1) ** 2,
+        dim=-1,
+    )
+    return entropy.item(), varentropy.item()
+
+
+def check_parameter_property(api_description, parameter_name, property_name):
```
adilhafeez, line 51: is_property_set?

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+        self.function_properties = {x["name"]: x["parameters"] for x in self.apis}
+
+    def process_token(self):
+        """
+        Processes the current token and updates the state and mask accordingly.
+        Detects hallucinations based on the token type and log probabilities.
+        """
+        content = "".join(self.tokens).replace(" ", "")
+        if self.current_token == "<tool_call>":
+            self.mask.append("t")
```
adilhafeez, line 122: consider using ENUM

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+        Returns:
+            int: The number of consecutive occurrences of the token.
+        """
+        return (
+            len(list(itertools.takewhile(lambda x: x == token, reversed(self.mask))))
+            if self.mask and self.mask[-1] == token
+            else 0
+        )
+
+    def check_function_name(self):
```
adilhafeez, line 219: check_function_hallucination

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+        Checks the extracted function name against the function descriptions.
+        Detects hallucinations if the function name is not found.
+        """
+        f_len = self.count_consecutive_token("f")
+        self.function_name = "".join(self.tokens[:-1][-f_len:])
+        if self.function_name not in self.function_description.keys():
+            self.hallucination = True
+            self.hallucination_message = f"Function name '{self.function_name}' not found in given function descriptions."
+
+    def check_parameter_name(self):
```
adilhafeez, line 230: check_param_hal...

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,240 @@
...
+        # Parameter name extraction logic
+        if self.state == "parameter_name" and not content.endswith(
+            const.PARAMETER_NAME_END_TOKENS
+        ):
+            self.mask.append("p")
+        elif self.state == "parameter_name" and content.endswith(
+            const.PARAMETER_NAME_END_TOKENS
+        ):
+            self.state = None
+            self.check_parameter_name()
```
adilhafeez, line 146: if hallucination is detected then break here

adilhafeez (commented):
https://medium.com/@amirm.lavasani/design-patterns-in-python-iterator-2d3e9917f930

model_server/app/function_calling/hallucination_handler.py
cotran2: how to handle when hallucinated will be in a different pr

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,292 @@
...
+        if self.response_iterator is not None:
+            try:
+                r = next(self.response_iterator)
+                if hasattr(r.choices[0].delta, "content"):
+                    token_content = r.choices[0].delta.content
+                    if token_content:
+                        logprobs = [
+                            p.logprob
+                            for p in r.choices[0].logprobs.content[0].top_logprobs
+                        ]
```
adilhafeez, line 149: ensure that logprobs exist

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,292 @@
...
+            raise ValueError("API descriptions not set.")
+        parameter_names = {}
+        for func in self.function:
+            func_name = func["name"]
+            parameters = func["parameters"]["properties"]
+            parameter_names[func_name] = list(parameters.keys())
+        self.function_description = parameter_names
+        self.function_properties = {x["name"]: x["parameters"] for x in self.function}
+
+    def check_token_hallucination(self, token, logprob):
```
adilhafeez, line 119: check_and_append*

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,292 @@
...
+            except StopIteration:
+                raise StopIteration
+
+    def process_token(self):
+        """
+        Processes the current token and updates the state and mask accordingly.
+        Detects hallucinations based on the token type and log probabilities.
+        """
+        content = "".join(self.tokens).replace(" ", "")
+        if self.current_token == "<tool_call>":
```
adilhafeez, line 161: "<tool_call>" = const

adilhafeez (commented):
internal func should be prefix by _

nehcgs:
@cotran2 Thanks for getting this up!

1. The new parameters defined in `constants.py` should be moved to the corresponding function as they're only used there.
2. Consider dropping the hallucination handler: 
    - Seems you want to handle all errors in the same place. In that case, you should separate non-hallucination stuff (e.g., function name and parameter name errors) out of there. Not all errors can be classified as hallucinations. If you want to work on it now, create a `ErrorHandler` to check all errors comprehensively (e.g., data type and format checks) and it will make it easier when we want to extend it with error checks on others.
    - It doesn't make sense that you maintain so many variables in `HallucinationStateHandler`. You can simply maintain a flag to indicate difference states so that you can take difference action accordingly. Remove everything not related to hallucination and try to compress it as a single function. 
    - For all the constants defined in `constants.py` but used here, think about a way to get these constants dynamically from the system prompt of the model. Otherwise, you will have to change them every time we change the system prompt.
    - This hallucination check can only be used for `Arch-Function`. It probably should be independent. You may want to put it within `function_calling/model_handler.py`. 
3. I'm not sure why you choose different set of thresholds for `<tool_call>` and parameter value checks. Both of them are from the same generation, so should be the same in my view. Please check whether your experiments overfit to them or not.
4. The default value check in `HallucinationStateHandler` is also necessary. It happens mostly because of the poor definition of functions. In that case, we should preprocess that.
5. The variable names `function_description ` and `function_properties ` are very ambiguous. I'm not sure if we still need them. If yes, try to give them different names.

cotran2:
To address @nehcgs
1.  I don't want to maintain every constants in each file. Those will be used in the future for unittests.
2. Because we process the stream response from OpenAI client at once so it makes sense to me to handle everything with the same object. I can change the errors type but if model produces wrong function_name/parameter name, those are not hallucination?
- I can try to compress but the variables are for debugging/making sure it works/intepretable as well.
Based on my experiment, <tool_call> threshold is lower than parameter value token threshold to maintain the same FPR.

nehcgs:
> To address @nehcgs
> 
> 1. I don't want to maintain every constants in each file. Those will be used in the future for unittests.
> 2. Because we process the stream response from OpenAI client at once so it makes sense to me to handle everything with the same object. I can change the errors type but if model produces wrong function_name/parameter name, those are not hallucination?
> 
> * I can try to compress but the variables are for debugging/making sure it works/intepretable as well.
>   Based on my experiment, <tool_call> threshold is lower than parameter value token threshold to maintain the same FPR.

@cotran2 
1. If you have to change the logic or design only for tests, then you should consider whether your tests are designed appropriately. For your current tests, you don't use any of these constants. If you need these constants, you may want to make a call to this function directly. Similar to what happened to system prompts, you define it twice in different tests, but you can simply call `function_calling/model_handler.py` to get them to avoid such long repeated code.
2. The point is, you maintain a lot of variables there. You have processed all of them, why still keep them? You're currently using the first token logprobs, so there's no dependency between tokens. You can simply drop everything you processed so that the space complexity can be as low as O(1).
3. If you take 2 into consideration, you won't have trouble in verifying so many variables. 
4. For `<tool_call>`, I think you may want to double-check to avoid overfitting. If you believe they should have different thresholds, then why you don't give different thresholds for different parameter values? You can always break down this problem and set thresholds for every parameter value to improve the performance, but we need to get findings to demonstrate this.

nehcgs:
@cotran2  Let's keep `HallucinationStateHandler` as it is. I will refactor this part later. Please address other issues and get this in. Everything else looks good to me.

model_server/app/function_calling/hallucination_handler.py
```diff
@@ -0,0 +1,324 @@
...
+                                for p in r.choices[0].logprobs.content[0].top_logprobs
+                            ]
+                        except Exception as e:
+                            raise ValueError(
+                                f"Error extracting logprobs from response: {e}"
+                            )
+                        self.append_and_check_token_hallucination(
+                            token_content, logprobs
+                        )
+                        return token_content
```
adilhafeez, line 181: good change

adilhafeez (approved):
@cotran2 I've reviewed from my side. I think code is in a state where we can iterate it and continue to improve it as we integrate in our code base.

But please get approval from @nehcgs as well.

PR #292, opened 2024-11-20
Remove blanket unused imports from the common crate

junr03:
Signed-off-by: José Ulises Niño Rivera <junr03@users.noreply.github.com>


adilhafeez (approved):
This is good cleanup. I thought clippy would catch this.

junr03:
> I thought clippy would catch this.

It does except that it was commanded not to by this [line](https://github.com/katanemo/archgw/pull/292/files#diff-84e11d13acaf28044921bb5a427cfbf511c70a0b3881949137ba4093df58e266L1)

PR #303, opened 2024-11-25
use fix prompt guards

### LINKED ISSUE #274: Inclusion of prompt_guards in arch_config results in...
"Error validating arch_config file ... ('prompt_guards' was unexpected) ...". This is because the arch_config_schema.yaml doesn't include any mention of prompt_guards in the schema. We house the schema in two places  https://github.com/katanemo/arch/blob/main/docs/source/resources/includes/arch_config_full_reference.yaml and https://github.com/katanemo/arch/blob/main/arch/arch_config_schema.yaml

We need to validate the schema we have published to our docs and make sure that all the supported features have test cases.

adilhafeez:
Updated code to accept prompt_guard and updated e2e tests to validate prompt_guards

fixes #274 

image

### DIFF
--- a/crates/prompt_gateway/src/stream_context.rs
+++ b/crates/prompt_gateway/src/stream_context.rs
@@ -1076,7 +1076,7 @@ impl StreamContext {
                 .unwrap_or("refrain from discussing jailbreaking.");
             info!("jailbreak detected: {}", msg);
 
-            let default_target_response_str = if self.streaming_response {
+            let response_str = if self.streaming_response {
                 let chunks = vec![
                     ChatCompletionStreamResponse::new(
                         None,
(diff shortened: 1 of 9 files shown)


crates/prompt_gateway/src/stream_context.rs
```diff
@@ -1074,7 +1074,36 @@ impl StreamContext {
...
+
+                to_server_events(chunks)
+            } else {
+                let chat_completion_response = ChatCompletionsResponse::new(msg.to_string());
+                serde_json::to_string(&chat_completion_response).unwrap()
+            };
+
+            self.send_http_response(
+                StatusCode::OK.as_u16().into(),
+                vec![("Powered-By", "Katanemo")],
```
salmanap, line 1103: We should remove this.

salmanap (changes requested):
Everything looks good. Just remove the unnecessary response header

salmanap (approved):

Issue #226, opened 2024-10-28
Don't use docker-compose for archgw


adilhafeez:
docker-compose file isn't needed to start arch gateway process since we don't have another core service we run through docker

PR #305, opened 2024-11-26
remove dependency on docker-compose when starting up archgw

adilhafeez:
fixes https://github.com/katanemo/archgw/issues/226


arch/tools/cli/core.py
```diff
@@ -1,40 +1,73 @@
...
+        ports={
+            "10000/tcp": 10000,
+            "10001/tcp": 10001,
+            "11000/tcp": 11000,
+            "12000/tcp": 12000,
+            "19901/tcp": 19901,
+        },
+        volumes={
+            f"{arch_config_file}": {
+                "bind": "/app/arch_config.yaml",
```
salmanap, line 38: does this require someone to checkout main to run archgw?

arch/tools/cli/core.py
```diff
@@ -1,40 +1,73 @@
...
 from huggingface_hub import snapshot_download
+from dotenv import dotenv_values
+
 
 log = getLogger(__name__)
 
+dot_env = dotenv_values(".env")
+
+
+def start_archgw_docker(client, arch_config_file):
```
salmanap, line 24: where does the CLI need to be to start arch?

arch/tools/build_cli.sh
```diff
@@ -1,23 +1,4 @@
 #!/bin/bash
 
-# Define paths
-source_schema="../arch_config_schema.yaml"
```
salmanap, line 4: Its' unclear to me how the config_generator.py gets access to this schema to validate the arch_config from the developer. We have to have the cli run independently - meaning it should not require a fork of main. For those who want to `archgw build` from source, its necessary for them to fork main and run locally.

salmanap (changes requested):
I like the removal of the docker compose dependency - but I have some other questions around the usage of arch_config_schema.yaml

arch/tools/build_cli.sh
adilhafeez: schema file is burned into docker image - see here https://github.com/katanemo/archgw/blob/main/arch/Dockerfile#L26

arch/tools/cli/core.py
adilhafeez: here https://github.com/katanemo/archgw/pull/305/files#diff-9fb218171a8bf47ce0a494ac2ccdf5052d2525f6a0223b0ab327e942b3c48c87R132

arch/tools/cli/core.py
adilhafeez: no, its part of archgw cli

salmanap (approved):

PR #329, opened 2024-12-05
322 add support for pydantic logfire for llm agent tracing

aayushwhiz:
This makes a new weather forecast demo that has tracing with logfire set up


aayushwhiz:
I still need to update readme and ensure that logfire key is a required environment variable, but all the configs are ready

adilhafeez:
who is passing `LOGFIRE_API_KEY` env var to otel-collector? I think you may need to add this var in docker-compose.

You can also ensure that env var is passed in correctly using [env var interpolation](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax),

`${LOGFIRE_API_KEY:$?please supply logfire api key}`

You can also check for this var in `run_demo.sh` script.

salmanap (commented):
Can't we have one weather_service and toggle the different observability providers via the shell script. This will make maintaining all these demos hard. Looks like all we really needed was setting up a slightly different local OTEL collector

salmanap (changes requested):
Overall the changes look good. But I think we can do a few more things to clean the refactor.

1. The weather_forecast demo should have main.py in it, along with the docker-compose files for each tracing option. You can move the otel_cellector folder into shared as that might be handy to other demos as well. 

2. the run_demo.sh should default to "docker-compose.yml' - you have it converging to docker-compose-jaeger at the moment. 

3. Update the README to talk about the project structure a bit (as there are multiple compose files). Of course the lead in should be what it is right now which is run the demo

demos/shared/logfire/otel-collector-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+      grpc:
+        endpoint: 0.0.0.0:4317
+      http:
+        endpoint: 0.0.0.0:4318
+
+exporters:
+  otlphttp:
+    endpoint: "https://logfire-api.pydantic.dev"
+    headers:
+      Authorization: "${LOGFIRE_API_KEY}"
```
salmanap, line 13: How does a developer set this logfire key? Do we take that in the run_demo.sh script?

demos/weather_forecast/README.md
```diff
@@ -1,27 +1,56 @@
...
+1. For Jaeger, you can either use the default run_demo.sh script or run the following command:
+
+```sh
+sh run_demo.sh jaeger
+```
+
+2. For Logfire, first make sure to add a LOGFIRE_API_KEY to the .env file. You can either use the default run_demo.sh script or run the following command:
+
+```sh
+sh run_demo.sh logfire
```
salmanap, line 43: we need to take in the developer key here so that we fail fast if the logfire dev key is not provided

demos/weather_forecast/docker-compose-jaeger.yaml
```diff
@@ -0,0 +1,41 @@
+services:
```
salmanap, line 1: I presume this file is copied, and did we test this after the new folder structure?

demos/weather_forecast/run_demo.sh
```diff
@@ -1,13 +1,36 @@
 #!/bin/bash
 set -e
 
+# Directory where the Docker Compose files are stored
+
+# Function to determine the docker-compose file based on the argument
+get_compose_file() {
+  case "$1" in
+    jaeger)
+      echo "docker-compose-jaeger.yaml"
```
salmanap, line 10: why are we echoing the value?

demos/weather_forecast/run_demo.sh
```diff
@@ -1,13 +1,36 @@
...
+      echo "docker-compose.yaml"  # Default to Jaeger
+      ;;
+  esac
+}
+
 # Function to start the demo
 start_demo() {
-  # Step 1: Check if .env file exists
+  # Step 1: Determine the docker-compose file
+  COMPOSE_FILE=$(get_compose_file "$1")
```
salmanap, line 27: Is this the correct pattern to return from a function in shell?

demos/weather_forecast/run_demo.sh
```diff
@@ -18,22 +41,26 @@ start_demo() {
...
 
 # Function to stop the demo
 stop_demo() {
-  # Step 1: Stop Docker Compose services
-  echo "Stopping Network Agent using Docker Compose..."
-  docker compose down
+  echo "Stopping all Docker Compose services..."
+
+  # Stop all services by iterating through all configurations
+  for compose_file in ./docker-compose*.yaml; do
```
salmanap, line 58: Why are we looping through these files? shouldn't we always have one compose file to shut down?

demos/weather_forecast/run_demo.sh
```diff
@@ -42,6 +69,6 @@ stop_demo() {
 if [ "$1" == "down" ]; then
   stop_demo
 else
-  # Default action is to bring the demo up
-  start_demo
+  # Use the argument (jaeger, logfire, signoz) to determine the compose file
```
salmanap, line 72: We need to add the developer_key argument for logfire here

salmanap (commented):
Why do we need the weather_forecast_service folder anymore? Why don't we fold its file in the weather_forecast folder? This way all the code for weather_forecast is in a single place and easy to navigate for the developers looking at our code base

demos/weather_forecast/docker-compose-jaeger.yaml
aayushwhiz: yes, this was tested after the new folder structure with jaeger ui and everything

demos/weather_forecast/run_demo.sh
aayushwhiz: Different compose files would shut down different things, so this was just a way to sort of ensure everything is shutdown

aayushwhiz: wait are you recommending having a compose file specific to shutdown?

demos/shared/logfire/otel-collector-config.yaml
aayushwhiz: no it is currently setup as an environment variable

demos/weather_forecast/run_demo.sh
salmanap: you only start up one docker-compose file. So why are you shutting down all of them?

salmanap (changes requested):
Please first mark open comments as resolved if true. And there are still open comments, so please address. One major one: I am not sure why we have the weather_forecast_service folder, when things have folded into weather_forecast

demos/weather_forecast/run_demo.sh
aayushwhiz: That is the way I saw online for how people were returning from the function, i adjusted it so that it does not output to the user's terminal

demos/weather_forecast/run_demo.sh
aayushwhiz: because when sh run_demo.sh down is called, there is no way to know which docker compose file was used

salmanap: Ask for a parameter, else shutdown the default compose file. This would be consistent with the start command. Agree we can do better here, but I won't block on that.

salmanap (commented):
One minor comment on stop. We should follow the pattern on run_demo.sh start

demos/weather_forecast/run_demo.sh
```diff
@@ -1,47 +1,95 @@
 #!/bin/bash
 set -e
 
+# Function to load environment variables from the .env file
+load_env() {
+  if [ -f ".env" ]; then
+    export $(grep -v '^#' .env | xargs)
+  fi
+}
```
adilhafeez, line 9: nice!

adilhafeez (approved):

PR #330, opened 2024-12-05
update getting started guide and add llm gateway and prompt gateway samples

adilhafeez:
This is how you can interact now

```bash
$ curl --header 'Content-Type: application/json' \
  --data '{"messages": [{"role": "user","content": "what is exchange rate for gbp"}]}' \
  http://localhost:10000/v1/chat/completions | jq ".choices[0].message.content"
"As of the date provided in your context, December 5, 2024, the exchange rate for GBP (British Pound) from USD (United States Dollar) is 0.78558. This means that 1 USD is equivalent to 0.78558 GBP."

```

Could also use chatbot ui,

image

This code change also adds changes,
- Add ability to use parameters in URL
- Add support for http GET method
- Add support for using https clusters


README.md
```diff
@@ -39,88 +39,79 @@ To get in touch with us, please join our [discord server](https://discord.gg/pGZ
 
 ## Quickstart
 
-Follow this guide to learn how to quickly set up Arch and integrate it into your generative AI applications.
+Follow this quickstart guide to use arch gateway to manage access keys, provide unified access to upstream LLMs and to provide e2e observability. Later we will see how you can use arch gateway to build a Gen AI application.
```
salmanap, line 42: We need to highlight the most differentiating feature first - build apps with Arch using simple APIs. Followed by observability, access key management and other features. We are uniquely positioned on and shouldn't bury the lede: https://www.merriam-webster.com/wordplay/bury-the-lede-versus-lead

README.md
```diff
@@ -39,88 +39,79 @@ To get in touch with us, please join our [discord server](https://discord.gg/pGZ
...
 
 ```console
 $ python -m venv venv
 $ source venv/bin/activate   # On Windows, use: venv\Scripts\activate
-$ pip install archgw
+$ pip install archgw==0.1.5
 ```
 
-### Step 2: Configure Arch with your application
+### Build LLM gateway
```
salmanap, line 65: not sure what "build LLM gateway" means. For developers its install Arch, configure Arch and run Arch. We should move just the routing (outbound) example to later in the README.

README.md
```diff
@@ -39,88 +39,79 @@ To get in touch with us, please join our [discord server](https://discord.gg/pGZ
...
-  app_server:
-    # value could be ip address or a hostname with port
-    # this could also be a list of endpoints for load balancing
-    # for example endpoint: [ ip1:port, ip2:port ]
-    endpoint: host.docker.internal:18083
-    # max time to wait for a connection to be established
-    connect_timeout: 0.005s
 ```
-### Step 3: Using OpenAI Client with Arch as an Egress Gateway
+$ archgw up arch_config.yaml
```
salmanap, line 102: This output could change and the README would look stale. Any reason this conveys value in the getting started section?

README.md
```diff
@@ -143,6 +134,161 @@ print("OpenAI Response:", response.choices[0].message.content)
 
 ```
 
+#### Step 3.2: Using curl command
```
salmanap, line 137: I like this usage pattern.

README.md
```diff
@@ -143,6 +134,161 @@ print("OpenAI Response:", response.choices[0].message.content)
...
+      },
+      ...
+    }
+  ],
+  ...
+}
+
+```
+
+### Build Gen AI Application
```
salmanap, line 188: Please move this up. Building with Arch should be the lead.

arch/tools/cli/config_generator.py
```diff
@@ -49,7 +49,9 @@ def validate_and_render_schema():
 
     if "prompt_targets" in config_yaml:
         for prompt_target in config_yaml["prompt_targets"]:
-            name = prompt_target.get("endpoint", {}).get("name", "")
+            name = prompt_target.get("endpoint", {}).get("name", None)
```
salmanap, line 52: Can you add a comments section on what this does? Does this remove the dependency on having an endpoint for a prompt_target?

crates/common/src/path.rs
```diff
@@ -0,0 +1,82 @@
...
+                result.push(c);
+            }
+        }
+    }
+
+    Ok(result)
+}
+
+#[cfg(test)]
+mod test {
```
salmanap, line 33: Excellent tests...

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -892,9 +893,37 @@ impl StreamContext {
         let endpoint = prompt_target.endpoint.unwrap();
         let path: String = endpoint.path.unwrap_or(String::from("/"));
 
+        // only add params that are of string, number and bool type
```
salmanap, line 896: this makes sense to me. We should add this in the documentation as well - so that its clear to developers what we swap into the parameters. Btw, how do we distinguish between string/number/bool - aren't they all represented as strings in the path?

demos/currency_exchange/arch_config.yaml
```diff
@@ -0,0 +1,50 @@
...
+prompt_targets:
+  - name: currency_exchange
+    description: Get currency exchange rate from USD to other currencies
+    parameters:
+      - name: currency_symbol
+        description: the currency that needs conversion
+        required: true
+        type: str
+        in_path: true
+    http_method: GET
```
salmanap, line 33: Should we default to GET or POST? Would make the developer experience easier for the common scenarios.

salmanap (changes requested):
I really like these changes, alot. My only major feedback is that we should lead with "Build with Arch" its the most magical and differentiating experience. We should lead with that first.

README.md
adilhafeez: This is just showing sample output of a command execution.

README.md
adilhafeez: moved

crates/common/src/path.rs
adilhafeez: chatgpt 😵‍💫

crates/prompt_gateway/src/stream_context.rs
adilhafeez: Other options are list (or array) and dict (objects). That wouldn't make sense for path param

demos/currency_exchange/arch_config.yaml
adilhafeez: GET should be default as its more common for simpler RESTful APIs.

adilhafeez: I will update the code

crates/prompt_gateway/src/stream_context.rs
adilhafeez: > how do we distinguish between string/number/bool

In yaml and in json string, number and bool are stored differently. For example,

```yaml
format: huggingface
```

here huggingface is a string.

Also this is same,

```yaml
format: "huggingface"
```

In following example 100 is a number and true is bool

```yaml
tracing:
  random_sampling: 100
  trace_arch_internal: true
```

demos/currency_exchange/arch_config.yaml
adilhafeez: updated the code

arch/tools/cli/config_generator.py
adilhafeez: this was unrelated code - removed it

adilhafeez: let me think a bit more over it

demos/insurance_agent/arch_config.yaml
```diff
@@ -26,6 +26,7 @@ endpoints:
 
 prompt_targets:
   - name: policy_qa
+    http_method: POST
```
salmanap, line 29: move in endpoint

salmanap: and update the reference spec in documentation.

demos/network_agent/arch_config.yaml
```diff
@@ -19,6 +19,7 @@ system_prompt: |
 prompt_targets:
     - name: device_summary
       description: Retrieve network statistics for specific devices within a time range
+      http_method: POST
```
salmanap, line 22: change code to GET

salmanap (commented):
Minor changes. Looks good.

demos/insurance_agent/arch_config.yaml
adilhafeez: moved

adilhafeez: updated reference

demos/network_agent/arch_config.yaml
adilhafeez: This isn't as simple as params need to be moved to path. Created issue  https://github.com/katanemo/archgw/issues/333

salmanap (approved):

PR #332, opened 2024-12-06
docs: update README.md

eltociear:
minor fix


salmanap (approved):

salmanap:
Thank you for catching this!

salmanap:
@adilhafeez we'll need to check if the arch-fc endpoint is down. this has happened a couple of times, so time to add some resiliency here.

PR #343, opened 2024-12-11
add more granular bucket sizes for ttft

aayushwhiz:
TTFT metric had really large bucket sizes and for the demos, most values were between 1000 and 2500 resulting in the prometheus metrics essentially only showing 1750 as the ttft rather than giving more detail, this PR changes the bucket sizes to work better for ttft metric.


aayushwhiz:
These are some screenshots of the stats page with this new change

Screenshot 2024-12-12 at 1 31 50 PM
Screenshot 2024-12-12 at 1 32 06 PM
Screenshot 2024-12-12 at 1 31 31 PM

adilhafeez (approved):

salmanap:
This is super cool. Thanks for pushing this out...

PR #346, opened 2024-12-18
Fix network agent demo

aayushwhiz:
the api_server container was not running properly so I updated the configs to fix the demo and made it run the shared chatbot ui.


adilhafeez:
pulled these changes in https://github.com/katanemo/archgw/pull/328

Issue #352, opened 2024-12-30
Update README to include Docker setup troubleshooting steps


salmanap:
https://github.com/katanemo/archgw/issues/349#issuecomment-2565203437 - see comment here.

Issue #351, opened 2024-12-30
Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))


salmanap:
This was reported in https://github.com/katanemo/archgw/issues/349 - and the issue stems from the fact that the archgw cli 0.1.7 uses a python docker client, which requires that the Docker desktop be configured with advanced settings. See screenshot.

This will cause other developers to trip up and make our setup more cumbersome. We should update the cli to not require this hard dependency. 

<img width="1271" alt="Screenshot 2024-12-30 at 12 04 02 PM" src="https://github.com/user-attachments/assets/6e05465b-0400-4879-ad73-7bd983e0ae72" />

PR #361, opened 2025-01-08
use per user docker socket if system docker socket doesn't exist

adilhafeez:
(venv) ➜  currency_exchange git:(main)  sh run_demo.sh
.env file already exists. Skipping creation.
...
2025-01-08 13:59:42,942 - cli.utils - INFO - Default docker socket /var/run/docker.sock not found, using unix:///Users/adilhafeez/.docker/run/docker.sock
...

fixes https://github.com/katanemo/archgw/issues/351
fixes https://github.com/katanemo/archgw/issues/352 (not needed anymore)

### DIFF
--- a/arch/tools/cli/utils.py
+++ b/arch/tools/cli/utils.py
@@ -1,3 +1,4 @@
+import os
 import yaml
 import logging
 import docker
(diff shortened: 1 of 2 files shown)


arch/tools/cli/utils.py
```diff
@@ -19,8 +20,26 @@ def getLogger(name="cli"):
 log = getLogger(__name__)
 
 
+def update_docker_host_env():
```
salmanap, line 23: Can we remove the dependency on the docker python client? Because this change will still require the developer to go to advanced settings and check of the box "Allow the default Docker socket to be used (requires password)" - which is work and generally hard to find.

adilhafeez: There is no change from docker desktop. This change will allow developer to connect to docker system without making any changes to docker system.

salmanap (approved):

PR #362, opened 2025-01-08
Some fixes on model server

nehcgs:
1. Improve logging
2. Clean up code and fix tests
3. A little bit refactor
4. Add simple formatting support


model_server/src/main.py
```diff
@@ -45,6 +43,13 @@
 )
 
 trace.get_tracer_provider().add_span_processor(BatchSpanProcessor(otlp_exporter))
+logging.getLogger("opentelemetry.exporter.otlp.proto.grpc.exporter").setLevel(
+    logging.ERROR
+)
```
adilhafeez, line 48: thanks!

model_server/src/main.py
```diff
@@ -62,73 +67,85 @@ async def models():
 
 @app.post("/function_calling")
 async def function_calling(req: ChatMessage, res: Response):
+    endpoint_logs = "=" * 30 + "  [Endpoint: /function_calling]  " + "=" * 30
```
adilhafeez, line 70: pls use logging.info

model_server/src/main.py
```diff
@@ -62,73 +67,85 @@ async def models():
...
+    endpoint_logs += f"\n  - [request body]: {json.dumps(req.model_dump())}"
+    logger.info(endpoint_logs)
+
+    final_response = ChatCompletionResponse()
```
adilhafeez, line 74: To declare var of certain type use following convention,
```
final_response: ChatCompletionResponse = None
```

model_server/src/main.py
```diff
@@ -62,73 +67,85 @@ async def models():
...
-        logger.error(f"Error in chat_completion /function_calling: {e}")
         res.status_code = 500
-        return {"error": f"[Arch-Intent] - {e}"}
+        error_messages = f"\n  - [Arch-Intent] - Error in ChatCompletion: {e}"
+
+    if error_messages is not None:
+        logger.error(error_messages)
+        final_response.metadata["error"] = error_messages
+
+    return final_response
```
adilhafeez, line 123: then here you can use sth similar to this,

```python
if not final_response:
    final_response = ChatCompletionResponse(
        metadata={
            "error": error_message
        },
        choices: [],
        model="--none--"
    )
```

adilhafeez (changes requested):
Shuguang - thanks for putting this PR together. I've some comments. Please let me know if you have any questions about them.

model_server/src/core/function_calling.py
```diff
@@ -160,7 +169,11 @@ class ArchFunctionConfig:
     TASK_PROMPT = textwrap.dedent(
         """
     You are a helpful assistant.
-    """
+
+    Today's date: {}
+    """.format(
+            utils.get_today_date()
+        )
```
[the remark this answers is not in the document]
nehcgs: This is for the support on date-related format. It won't affect results.

model_server/src/main.py
nehcgs: This will potentially return a None, which is not consistent with other outputs.

adilhafeez: Look at my comment toward the end of the function. Also with this change you don't need to set default values for choices and model in ChatCompletionResponse class

model_server/src/core/function_calling.py
```diff
@@ -189,7 +202,7 @@ class ArchFunctionConfig:
         "temperature": 0.6,
         "top_p": 1.0,
         "top_k": 10,
-        "max_tokens": 512,
+        "max_tokens": 1024,
```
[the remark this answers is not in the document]
nehcgs: For RAG scenarios, I think we need it.

model_server/src/commons/utils.py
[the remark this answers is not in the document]
adilhafeez: not blocker but pls consider this change

adilhafeez (changes requested):
Thanks for the update, left some comments around logging format. Please take a look at it and let me know if you have any questions.

model_server/src/core/function_calling.py
```diff
@@ -123,20 +123,20 @@ async def chat_completion(self, req: ChatMessage) -> ChatCompletionResponse:
...
+            logger.info(
+                "  - [Info]: No tools found, return `No` as the model response."
             )
         else:
             messages = self._process_messages(
                 req.messages, req.tools, self.extra_instruction
             )
 
-            handler_logs += f"\n  - [request]: {json.dumps(messages)}"
+            logger.info(f"  - [request]: {json.dumps(messages)}")
```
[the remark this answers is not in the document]
nehcgs: All the data is necessary. Trimming it will have a negative impact on debugging. This is the input to the endpoint, and we should keep it.

model_server/src/commons/utils.py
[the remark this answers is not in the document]
nehcgs: can you elaborate more on this?

adilhafeez (approved):

PR #370, opened 2025-01-17
[ENH] Enable LTO and Set codegen-units to 1 for Release Builds

### LINKED ISSUE #360: Enable Link-Time Optimization (LTO) and codegen-units = 1
Hi!

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance. If you want to read more about LTO and its possible modes, I recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I think you can enable LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project since LTO consumes an additional amount of time to finish the compilation routine. If you think that a regular Release build should not be affected by such a change as well, then I suggest adding an additional `dist` or `release-lto` profile where in addition to regular `release` optimizations LTO will also be added. Such a change simplifies life for maintainers and others interested in the project persons who want to build the most optimized version of the application. However, if we enable it on the Cargo profile level for the Release profile, users, who install the application with `cargo install` will get the LTO-optimized version of the game "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48). You also could be interested in other optimization options like `codegen-units = 1` - this option also brings performance improvements to the project.

Basically, it can be enabled with the following lines:

```
[profile.release]
codegen-units = 1
lto = true
```

I have made quick tests (Fedora 41, Rust 1.83, the latest version of the project at the moment) - here are the results:

`libllm_gateway.so`:

* Release (current default): 11 Mib
* Release + `codegen-units = 1` + Fat LTO: 9.8 Mib

`libprompt_gateway.so`:

* Release (current default): 1.7 Mib
* Release + `codegen-units = 1` + Fat LTO: 1.4 Mib

Thank you.

samadpls:
Fixes #360

### DIFF
--- a/crates/Cargo.toml
+++ b/crates/Cargo.toml
@@ -1,3 +1,7 @@
 [workspace]
 resolver = "2"
 members = ["llm_gateway", "prompt_gateway", "common"]
+
+[profile.release]
+codegen-units = 1
+lto = true


samadpls:
Hi @adilhafeez , could you please re-run the workflow for this PR at your convenience? This helps me update the PR accordingly.

samadpls:
Got it, thanks! Sorry for the delay—got a bit busy. I'll update the PR once I have some time to work on it. Will let you know when it's ready."

PR #379, opened 2025-01-23
log improvements and some code refactor

adilhafeez:
Logs with this change,

```sh
[2025-01-30 23:00:20.052][40][debug][wasm] ... request received: llm provider hint: None, selected llm: gpt-4o
[2025-01-30 23:00:20.052][40][debug][wasm] ... sending request to model server
[2025-01-30 23:00:20.590][40][debug][wasm] ... model server response received
[2025-01-30 23:00:20.590][40][debug][wasm] ... dispatching api call to developer endpoint: frankfurther_api, path: /v1/latest?base=USD&symbols=JPY
[2025-01-30 23:00:21.037][40][debug][wasm] ... api call response received: status code: 200
[2025-01-30 23:00:21.037][40][debug][wasm] ... sending request to upstream llm
[2025-01-30 23:00:22.124][40][debug][wasm] ... time to first token: 2072ms
[2025-01-30 23:00:24.294][40][debug][wasm] ... request latency: 4243ms
```


crates/common/src/routing.rs
```diff
@@ -35,11 +34,9 @@ pub fn get_llm_provider(
     }
 
     if llm_providers.default().is_some() {
-        debug!("no llm provider found for hint, using default llm provider");
```
salmanap, line 38: How are you thinking about debug vs. trace logs. Or better, yet how should developers think about it so that they have a consistent view of those log levels throughout the system. I like that we are gutting out some of our debug logs, but would be curious to get that definition in our documentation as well.

crates/prompt_gateway/src/http_context.rs
```diff
@@ -86,7 +79,7 @@ impl HttpContext for StreamContext {
         };
 
         debug!(
-            "developer => archgw: {}",
+            "developer => archgw request body: {}",
```
salmanap, line 82: I don't think this format "developer => archgw" is quite effective in tailing our logs. Do we have an envoy pattern that we could leverage so that its easier to reason about the the sequence diagram of the request from clear text labels in the log lines that get printed locally in the CLI.

crates/prompt_gateway/src/http_context.rs
```diff
@@ -159,7 +152,7 @@ impl HttpContext for StreamContext {
             }
         };
 
-        debug!("archgw => archfc: {}", json_data);
+        debug!("archgw => modelserver request body: {}", json_data);
```
salmanap, line 155: archgw is the product. So if we want to componentize the logs then I would think about archgw_ingress => archgw_modelserver. But please look at comment above on log formats to make them both human and machine readable.

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -344,7 +344,7 @@ impl StreamContext {
         );
 
         debug!(
-            "archgw => api call, endpoint: {}{}, body: {}",
+            "archgw => developer api call endpoint: {}, path: {}, body: {}",
```
salmanap, line 347: Please see comment above - we have an opportunity to simplify the readable format of the logs so that developers can easily reason about the network sequence of the request path. I would (again) encourage you to think about the logs the envoy generates for egress/ingress traffic and be inspired by those. We don't have to copy that structure, but create for familiarity so that its easy for developers

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -442,25 +441,39 @@ impl StreamContext {
         self.resume_http_request();
     }
 
-    fn filter_out_arch_messages(&mut self, callout_context: &StreamCallContext) -> Vec<Message> {
+    fn get_system_prompt(&self, prompt_target: Option<PromptTarget>) -> Option<String> {
```
salmanap, line 444: Are we stuffing two different changes into this logs PR? Should we create a separate PR here? I am okay with either, just wanted to know more about what exactly is the inteded behaviro

salmanap (commented):
I really like the clean-up, but have some comments that require your attention.

crates/prompt_gateway/src/stream_context.rs
adilhafeez: PR description states that this change includes both log changes and bit of a code refactor. But I do think its good to keep PRs simple.

crates/prompt_gateway/src/http_context.rs
adilhafeez: its not effective to dump all the body, changing it to trace

crates/prompt_gateway/src/stream_context.rs
adilhafeez: changed it to trace

crates/common/src/routing.rs
adilhafeez: removed these extra logs and moved all request/response details into trace level logging

salmanap (approved):

PR #382, opened 2025-01-24
Fix llm_routing provider element

adilhafeez:
We replaced provider with provider_interface to make it more clear to developers about provider api/backend being used. During that upgrade we removed support for mistral in provider to encourage developers to start using provider_interface. But this demo was not updated to use provider_interface as it was using mistral. This code change fixes it by replacing provider with provider_interface.


demos/currency_exchange/arch_config.yaml
```diff
@@ -7,9 +7,9 @@ listener:
   connect_timeout: 0.005s
 
 llm_providers:
-  - name: gpt-4o
+  - name: openai-gpt-4o
```
adilhafeez, line 10:
```suggestion
  - name: gpt-4o
```

salmanap (approved):

salmanap (approved):

PR #386, opened 2025-01-29
add ability to specify custom http headers in api endpoint

adilhafeez:
Add support for specifying http headers in http request to api call endpoint.

For example in the following example we are setting auth header in http request,

```
...
  - name: stock_quote
    description: get current stock exchange rate for a given symbol
...
    endpoint:
      name: twelvedata_api
      path: /quote?symbol={symbol}
      http_headers:
        Authorization: "apikey $TWELVEDATA_API_KEY"
...
```

Note this change is not just limited to auth header you can use `http_headers` field to attach any arbitrary header to the request. For example in the following request we can request api endpoint to return json,

```
...
    endpoint:
      name: twelvedata_api
      path: /quote?symbol={symbol}
      http_headers:
        Accept: application/json
...
```

For a complete example see stock_exchange demo.


salmanap (approved):

demos/stock_exchange/arch_config.yaml
```diff
@@ -0,0 +1,69 @@
...
+        type: str
+        in_path: true
+      # - name: interval
+      #   description: Time interval
+      #   default: 1h
+      #   type: str
+      #   in_path: true
+    endpoint:
+      name: twelvedata_api
+      path: /time_series?symbol={symbol}&interval=1day
```
salmanap, line 61: We should move query params out of the path. Arch-Function is senstive to these scenarios and if we highlight this as a common use case, we will see developers trip up in other scenarios. Please make them parameters and remove in_path for this scenario.

salmanap (commented):
One minor comment

demos/stock_exchange/arch_config.yaml
adilhafeez: trying it out

adilhafeez: updated

Issue #405, opened 2025-02-12
Bug: Additional properties are not allowed ('stream' was unexpected)


SalAlba:
This's not working.

Error validating arch_config file: /app/arch_config.yaml, schema file: arch_config_schema.yaml, error: Additional properties are not allowed ('stream' was unexpected)

check the link:
https://github.com/katanemo/archgw/blob/28ee14a75cbb06310e0da03b61eba7789b71a7ed/docs/source/guides/includes/arch_config.yaml#L16

``` shell

2025-02-12 11:50:12,016 - cli.main - INFO - Starting archgw cli version: 0.2.1
2025-02-12 11:50:12,017 - cli.main - INFO - Validating /home/sal/workspace/llm-rag-archgw-arch/arch_config.yaml
2025-02-12 11:50:13,235 - cli.main - INFO - Exiting archgw up: validation failed
2025-02-12 11:50:13,235 - cli.main - INFO - Error: Validation failed. Container exited with code {'StatusCode': 1}.
Logs:
Error validating arch_config file: /app/arch_config.yaml, schema file: arch_config_schema.yaml, error: Additional properties are not allowed ('stream' was unexpected)
Additional properties are not allowed ('stream' was unexpected)

Failed validating 'additionalProperties' in schema['properties']['llm_providers']['items']:
    {'type': 'object',
     'properties': {'name': {'type': 'string'},
                    'provider': {'type': 'string', 'enum': ['openai']},
                    'provider_interface': {'type': 'string',
                                           'enum': ['openai', 'mistral']},
                    'access_key': {'type': 'string'},
                    'model': {'type': 'string'},
                    'default': {'type': 'boolean'},
                    'endpoint': {'type': 'string'},
                    'protocol': {'type': 'string',
                                 'enum': ['http', 'https']},
                    'http_host': {'type': 'string'}},
     'additionalProperties': False,
     'required': ['name', 'model']}

On instance['llm_providers'][0]:
    {'name': 'dumy',
     'default': True,
     'stream': False,
     'provider_interface': 'openai',
     'access_key': '$AZURE_OPENAI_API_KEY',
     'model': '$AZURE_OPENAI_DEPLOYMENT_NAME',
     'endpoint': '172.17.0.3:3030'}
```

salmanap:
Sorry about this. Looking

adilhafeez:
looking

salmanap:
[...]

PR #407, opened 2025-02-12
Update arch_config and add tests for arch config file

adilhafeez:
- revamp listeners config section
new listeners section
```
listeners:
  ingress_traffic:
    address: 0.0.0.0
    port: 10000
    message_format: openai
    timeout: 5s
  egress_traffic:
    address: 0.0.0.0
    port: 12000

```
old section
```
listener:
  address: 0.0.0.0 # or 127.0.0.1
  port: 10000
  # Defines how Arch should parse the content from application/json or text/pain Content-type in the http request
  message_format: huggingface
  common_tls_context: # If you configure port 443, you'll need to update the listener with your TLS certificates
    tls_certificates:
      - certificate_chain:
          filename: /etc/certs/cert.pem
        private_key:
          filename: /etc/certs/key.pem
```
- allow default param types to be of int, bool or str
for example.
```
    parameters:
      - name: device_id
        type: int
        description: Identifier of the network device to reboot.
        default: 5
```
- allow default enum types in param list to be of int, bool or str
for example
```
      - name: confirmation
        type: bool
        description: Confirmation flag to proceed with reboot.
        default: false
        enum: [true, false]
```
- set default timeout for model server and api call to be 30s (TODO add ability to override these using config)
- update all config files to be compliant with new settings
- add github action test to validate all arch_config files

fixes https://github.com/katanemo/archgw/issues/405

here is a sample trace for currency converter demo

image


arch/arch_config_schema.yaml
```diff
@@ -3,21 +3,38 @@ type: object
...
+            type: string
+        additionalProperties: false
+      llm_gateway:
+        type: object
+        properties:
+          address:
+            type: string
+          port:
+            type: integer
+          message_format:
```
salmanap, line 30: Do we need the messages_format here?

arch/arch_config_schema.yaml
```diff
@@ -3,21 +3,38 @@ type: object
...
-        type: integer
-      message_format:
-        type: string
-      connect_timeout:
-        type: string
-    additionalProperties: false
-    required:
-      - address
-      - port
+      prompt_gateway:
```
salmanap, line 9: We describe ourselves as an edge and LLM proxy for agentic apps. A single gateway but capabilities to handle incoming prompt traffic to application backed and outgoing traffic to LLMs. While its tempting to qualify them as _gateway. We should keep our terminology consistent with the industry and easy to follow

ingress: Traffic: Traffic entering a network or system (proxy acting as an entry point).
egress: Traffic leaving a network or system (proxy acting as an outbound gateway).

salmanap: Note enoy doesn't have first class primitives for ingress/egress traffic as that behavior is handled via filter chains. So we depart a bit from the terms/concepts of envoy. But still in line with how an average developer might describe what we are doing. 

I would equally be okay with terms like

edge
forward

arch/arch_config_schema.yaml
```diff
@@ -107,15 +124,21 @@ properties:
               required:
                 type: boolean
               default:
-                type: string
+                anyOf:
```
salmanap, line 127: We should test if DeepSeek and AzureOpenAI endpoints now work with this change

arch/envoy.template.yaml
```diff
@@ -443,7 +441,7 @@ static_resources:
 
   clusters:
     - name: openai
-      connect_timeout: 5s
+      connect_timeout: 0.5s
```
salmanap, line 444: Any particular reason we are lowering these values?

arch/tools/cli/config_generator.py
```diff
@@ -104,7 +104,25 @@ def validate_and_render_schema():
     arch_config_string = yaml.dump(config_yaml)
     arch_llm_config_string = yaml.dump(config_yaml)
 
+    prompt_gateway_listener = config_yaml.get("listeners", {}).get("prompt_gateway", {})
```
salmanap, line 107: Lets make these optional, and default to 10000 for ingress/ege and 12000 for egress/forward

arch/validate_arch_config.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+failed_files=()
```
salmanap, line 3: Great change!

salmanap (commented):
Overall this is a great improvement. I do think we shouldn't introduce net new terms that would be hard for developers to follow. I would recommend for incoming traffic, we describe it as edge/ingress/inbound and for outbound traffic we describe it as egress/forward/outbound

Separately, I think we should think about sensible defaults so that the config file doesn't need to be verbose. If the listeners' section is added then we honor the traffic direction that developers want us to handle/process.

arch/arch_config_schema.yaml
adilhafeez: this is for schema validation - we need it here to enforce type of message received. Today we accept openai tomorrow we could accept claude natively for example.

arch/arch_config_schema.yaml
adilhafeez: using ingress/egress

arch/envoy.template.yaml
adilhafeez: 5s was quite high to start with which isn't right

arch/tools/cli/config_generator.py
adilhafeez: we do have 10000 and 12000 as default, look at few lines below

salmanap (approved):
LGTM. Ship it

PR #412, opened 2025-02-13
use docker cli to communicate to docker sub system

adilhafeez:
This code change removes dependency on docker http host. And after this change we will be compatible with podman.


salmanap (commented):
Does this change fix the issue we had with having a separate build_filter_image.sh script. Please comment and I am still reviewing.

adilhafeez:
> Does this change fix the issue we had with having a separate build_filter_image.sh script. Please comment and I am still reviewing.

That is totally separate thing. I use that in github action script to build docker image. But we may not need this file. I will take care of it in separate PR.

salmanap (approved):

PR #414, opened 2025-02-20
add k8s demo for api server

adilhafeez:


salmanap (commented):
Please move this to use cases. This isn't a sample python application. Also please help me understand the demo a bit better.

demos/use_cases/k8s_api/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Summary
+
+In this demo, you will learn how to set up and run a model server, Arch Gateway, and a Docker container for the UI. You will also learn how to tail access logs for debugging purposes. The demo includes examples of getting namespace details and retrieving pod information from a Kubernetes cluster.
+
+# How to run this demo.
+
+Before starting make sure you have completed the pre-requisites [here](https://github.com/katanemo/archgw?tab=readme-ov-file#prerequisites)
+
+In separate window start 1) model server, 2) arch gateway 3) docker container for UI and for debugging 4) tail access logs
```
salmanap, line 13: model_server is our internal implementation detail - while they can run this in foreground. I am not sure why this example forces developers to run the demo using these commands. Someone looking at this demo will be confused between arch gateway and model_server. 

Instead we should follow the pattern in all other demos. We could have an advanced "debug logs" demo

demos/use_cases/k8s_api/README.md
```diff
@@ -0,0 +1,44 @@
+# Kubernetes API Demo
+
+This demo shows how you can use a publicly hosted REST API and interact with it using the Arch Gateway.
```
salmanap, line 3: We should be more specific about the demo - interact with live APIs like the K8s control plane APIs using Arch Gateway

salmanap (changes requested):
minor comments.

PR #432, opened 2025-03-12
add preliminary support for llm agents

adilhafeez:


crates/prompt_gateway/src/stream_context.rs
```diff
@@ -312,12 +315,59 @@ impl StreamContext {
...
+
+                if let Some(overrides) = self.overrides.as_ref() {
+                    if overrides.use_agent_orchestrator.unwrap_or_default() {
+                        metadata.insert("use_agent_orchestrator".to_string(), "true".to_string());
+                    }
+                }
+
+                let messages = self.construct_llm_messages(&callout_context);
+
+                let chat_completion_request = ChatCompletionsRequest {
```
salmanap, line 346: for agent orchestrator we presume that we have to send to /v1/chat/completions for the agent itself? I think the simplification was that we would process SSE from the agent but why impose a strict path on the agent itself?

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -441,6 +499,40 @@ impl StreamContext {
             }
         };
 
+        if !prompt_target
```
salmanap, line 502: This is essentially adding streaming from the api server? how would this work? I thought that we couldn't stream from the tool_call_response because envoy had already opened a connection to an upstream LLM. Can you please clarify?

demos/use_cases/orchestrating_agents/arch_config.yaml
```diff
@@ -0,0 +1,47 @@
...
+    port: 12000
+    message_format: openai
+    timeout: 30s
+
+overrides:
+  use_agent_orchestrator: true
+
+endpoints:
+  agent_gateway:
+    agent_orchestrator: true
```
salmanap, line 21: I am not sure i follow use_agent_orchestrator (in overrides) and agent_orchestrator in endpoints. What's the difference and what's the use case? Should the developer simply want us to do agent orchestration without having to specify it in two different places? 

Please clarify.

demos/use_cases/orchestrating_agents/main.py
```diff
@@ -0,0 +1,150 @@
...
+tracer = trace.get_tracer(__name__)
+
+logger = logging.getLogger("uvicorn.error")
+logger.setLevel(logging.INFO)
+
+app = FastAPI()
+FastAPIInstrumentor().instrument_app(app)
+
+# Configure the OTLP exporter (Jaeger, Zipkin, etc.)
+otlp_exporter = OTLPSpanExporter(
```
salmanap, line 37: Where is this being used? Is it to connect the traceparent from the ingress envoy to the egress portion? Where are we adding any more instrumentation in code?

model_server/src/core/function_calling.py
```diff
@@ -533,52 +550,62 @@ async def chat_completion(self, req: ChatMessage) -> ChatCompletionResponse:
...
 
         if extracted["status"]:
             # Response with tool calls
             if len(extracted["result"]):
-                verified = self._verify_tool_calls(
-                    tools=req.tools, tool_calls=extracted["result"]
-                )
+                verified = {}
+                if use_agent_orchestrator:
+                    verified = {"status": True, "message": ""}
```
salmanap, line 604: I think we should put a comment here. This isn't done. Essentially for this path we will have to check for hallucinations as well

salmanap (commented):
I have a few clarifying questions and a few comments.  Lets chat and sort this out. We should put a clear banner that agent orchestration is beta. I'll cover this in the docs.

crates/llm_gateway/src/stream_context.rs
```diff
@@ -270,10 +299,16 @@ impl HttpContext for StreamContext {
             .last()
             .cloned();
 
-        // override model name from the llm provider
-        deserialized_body
-            .model
-            .clone_from(&self.llm_provider.as_ref().unwrap().model);
+        let model_name = match self.llm_provider.as_ref() {
+            Some(llm_provider) => match llm_provider.model.as_ref() {
+                Some(model) => model,
```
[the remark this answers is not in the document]
salmanap: Resolving based on the following issue: https://github.com/katanemo/archgw/issues/440

demos/use_cases/orchestrating_agents/main.py
```diff
@@ -0,0 +1,105 @@
...
+    api_key="None",  # archgw picks the API key from the config file
+    base_url="http://host.docker.internal:12000/v1",
+)
+
+
+def call_openai(messages: List[Dict[str, str]], stream: bool):
+    completion = openai_client.chat.completions.create(
+        model="None",  # archgw picks the default LLM configured in the config file
+        messages=messages,
+        stream=stream,
```
salmanap, line 38: adding issue for connecting trace headers for incoming and outgoing requests.

PR #488, opened 2025-05-27
trim conversation if it exceed max limit of what router model can handle

adilhafeez:
In this code change trimming happens where we keep most recent messages and start trimming old conversation until total message token size is below max token size supported by router model.

One exception is that if we are down to single message then we will keep the conversation even if it exceed the max token count.

### DIFF
--- a/crates/brightstaff/src/lib.rs
+++ b/crates/brightstaff/src/lib.rs
@@ -1,2 +1,3 @@
 pub mod handlers;
 pub mod router;
+pub mod utils;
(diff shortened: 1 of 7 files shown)


cotran2 (commented):
Can we trigger this behavior only when the model is arch router? @adilhafeez

adilhafeez:
> Can we trigger this behavior only when the model is arch router? @adilhafeez

@cotran2 yes 100% actually this implementation of the trait is for arch-router. Idea of having trait Router was so we can have different implementation of it.

PR #491, opened 2025-05-29
separate out currency exchange and preference based routing

adilhafeez:


crates/brightstaff/src/main.rs
```diff
@@ -89,6 +89,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
 
         let llm_providers = llm_providers.clone();
         let service = service_fn(move |req| {
+            info!("main.rs: request headers: {:?}", req.headers());
```
adilhafeez, line 92:
```suggestion
```

PR #510, opened 2025-06-24
Add support for updating model preferences

adilhafeez:


salmanap:
I don't quite follow the model/usage pattern here. Are we taking a usage pattern and an actual LLM for preferences? if we need a name why don't we generate a psuedo random one. I am not sure what "code_generation" as a model means in this context

adilhafeez:
So the `model` in request represents model name definition in arch_config.yaml file. The example in sample code may not be great. But here is better example of model usage,

```
[
  {
    "model": "claude/claude-sonnet-4-0",
    "usage": "generating new code snippets, functions, or boilerplate based on user prompts or requirements"
  },
  {
    "model": "openai/gpt-4o",
    "usage": "understand and explain existing code snippets, functions, or libraries"
  }
]
```

PR #517, opened 2025-07-07
better model names

adilhafeez:
- use provider_name / model_name in model info endpoint
- in brightstaff use renderred arch_config file
- infer llm_provier name if not specified
- add tests for config_generator
- update arch_config validator to ensure that checked in rendered files are not diverging
- update corresponding rst and md files with new schema


salmanap (commented):
Two things

1/ If we are taking in the model name as prefix, then we need to test this
2/ Do we need the provider_interface now? Or the provider_interface is an optional field indicating a different interface?

salmanap (approved):
Okay with changes. But I would suggest to follow the industry pattern for provider/model vs. having two separate fields.

demos/use_cases/ollama/arch_config.yaml
```diff
@@ -9,10 +9,8 @@ listeners:
 
 llm_providers:
 
-  - name: local-llama
-    provider_interface: openai
-    model: llama3.2
-    endpoint: host.docker.internal:11434
+  - model: openai/llama3.2
```
salmanap: we should just call this my_llm_provider/llama3.2. and add provider_interface: openai

demos/use_cases/preference_based_routing/arch_config_rendered.yaml
```diff
@@ -0,0 +1,29 @@
+listeners:
```
salmanap: I don't think we should expose our internal implementation of a rendered config file - will confuse developers

docs/source/guides/llm_router.rst
```diff
@@ -85,29 +82,22 @@ Below is an example to show how to set up a prompt target for the Arch Router:
...
-        model: gpt-4o
-        usage: Generating new code snippets, functions, or boilerplate based on user prompts or requirements
-
-    - name: code_understanding
-        provider_interface: openai
-        access_key: $OPENAI_API_KEY
-        model: gpt-4.1
-        usage: understand and explain existing code snippets, functions, or libraries
 
+    - model: openai/gpt-4o-mini
```
salmanap, line 86: I think we need a place to describe the construction of the model name [provider_hint]/[model_id]

docs/source/resources/includes/arch_config_full_reference_rendered.yaml
```diff
@@ -0,0 +1,95 @@
+version: v0.1
```
salmanap: We should work backwards from the full reference file, not a rendered file as that could change and represents our internal implementation

demos/use_cases/preference_based_routing/arch_config_rendered.yaml
adilhafeez: removed

demos/use_cases/ollama/arch_config.yaml
adilhafeez: updated

docs/source/resources/includes/arch_config_full_reference_rendered.yaml
adilhafeez: we need rendered file - fixed it by ensuring that both rendered file and checked in file are same using github tests so we don't have two diverging files

docs/source/guides/llm_router.rst
adilhafeez: we do emit a failure where we specify provider/model_id usage pattern,

https://github.com/katanemo/archgw/pull/517/files#diff-7ab8d088b9a0315de5aab050606272c33ffadd37c02dadebd1090bc6dfe0ac6fR119-R122

```
Invalid model name {model_name}. Please provide model name in the format <provider>/<model_id>.
```

salmanap (approved):

PR #531, opened 2025-07-16
pass model name in header when a route is selected when using usage p…

adilhafeez:
…references

### DIFF
--- a/crates/brightstaff/src/handlers/chat_completions.rs
+++ b/crates/brightstaff/src/handlers/chat_completions.rs
@@ -104,7 +104,7 @@ pub async fn chat_completions(
 
     debug!("usage preferences from request: {:?}", usage_preferences);
 
-    let mut determined_route = match router_service
+    let determined_model = match router_service
         .determine_route(
             &chat_completion_request.messages,
             trace_parent.clone(),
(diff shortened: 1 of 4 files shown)


crates/brightstaff/src/handlers/chat_completions.rs
```diff
@@ -104,7 +104,7 @@ pub async fn chat_completions(
 
     debug!("usage preferences from request: {:?}", usage_preferences);
 
-    let mut determined_route = match router_service
+    let determined_model = match router_service
```
salmanap, line 107: I think we'd want to be clear here on what the router_service is returning. I believe its the model selected and the route detected. We should fix the naming

salmanap (approved):
minor comments

crates/brightstaff/src/router/llm_router.rs
```diff
@@ -151,21 +156,22 @@ impl RouterService {
...
-                route_name,
+                parsed_response,
                 router_response_time.as_millis()
             );
 
-            if let Some(ref route) = route_name {
-                if route == "other" {
-                    return Ok(None);
-                }
+            if let Some(ref route) = parsed_response {
```
salmanap, line 169: what happens if no route is found? How are we defaulting now?

salmanap (commented):
one more..

crates/brightstaff/src/router/llm_router.rs
adilhafeez: if parsed_response is None then we send back None. Which in turn results in selecting model from chat completion request

PR #533, opened 2025-07-19
In request path use same format for usage preferences as arch_config

adilhafeez:


crates/brightstaff/src/router/router_model_v1.rs
```diff
@@ -155,9 +155,14 @@ impl RouterModel for RouterModelV1 {
             .map(|prefs| {
                 let llm_route: Vec<RoutingPreference> = prefs
                     .iter()
-                    .map(|pref| RoutingPreference {
-                        name: pref.name.clone(),
-                        description: pref.usage.clone().unwrap_or_default(),
+                    .flat_map(|pref| {
```
salmanap: we do this flat_map function twice (one via usage_preferences and other when we load the config). It might be helpful to put this in a helper function. Else, you'll have to update/maintain this thing in two places.

adilhafeez: Let me add some comments here to explain whats going on. There is a lot going on here. Simplifying it.

adilhafeez: refactored code

PR #537, opened 2025-07-23
update instructions on running arch-router locally

adilhafeez:


salmanap:
https://docs.archgw.com/concepts/llm_providers/supported_providers.html#ollama

Issue #535, opened 2025-07-22
Image content type failing in chat completion request


adilhafeez:
Customer reported that image content in chat completion request is not going through as expected. Here is the error they are seeing,

```
Input should be 'file' [type=literal_error, input_value='image_url', input_type=str]
```

Link to reported issue: https://discord.com/channels/1292630766827737088/1292630768283029638/1397199184267968515

PR #539, opened 2025-07-24
bug fix - allow image content to pass through

adilhafeez:
fixes https://github.com/katanemo/archgw/issues/535

### DIFF
--- a/crates/hermesllm/src/providers/openai/types.rs
+++ b/crates/hermesllm/src/providers/openai/types.rs
@@ -36,8 +36,15 @@ pub enum MultiPartContentType {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ImageUrl {
+    pub url: String,
+}
+
+#[skip_serializing_none]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct MultiPartContent {
     pub text: Option<String>,
+    pub image_url: Option<ImageUrl>,
     #[serde(rename = "type")]
     pub content_type: MultiPartContentType,
 }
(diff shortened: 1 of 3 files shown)


crates/hermesllm/src/providers/openai/types.rs
```diff
@@ -35,9 +35,16 @@ pub enum MultiPartContentType {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ImageUrl {
+    pub url: String,
+}
+
+#[skip_serializing_none]
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct MultiPartContent {
```
salmanap, line 45: This looks good - but can we validate two things a) have we comprehensively implemented all the input fields that openai models take so that we can passthrough cleanly and b) how would this impact other providers?  meaning would we have to support variants for claude that might also take an image_url? 

Lastly, does this impact the messages we send to Arch-Router, and the preference-aligned work we do in RouterServicev1?

salmanap (commented):
One minor comment

crates/hermesllm/src/providers/openai/types.rs
adilhafeez: @salmanap inline

> have we comprehensively implemented all the input fields that openai models take so that we can passthrough cleanly

No we have not. I am sure there will be fields that are part of openai but they are getting dropped. We should warn if any of fields are getting dropped so that developer knows it. Created a separate issue for this - https://github.com/katanemo/archgw/issues/540

> b) how would this impact other providers? meaning would we have to support variants for claude that might also take an image_url?

Other providers will have issues and may not work. There are slight differences between how to embed images between different providers. Issue filed https://github.com/katanemo/archgw/issues/541

> Lastly, does this impact the messages we send to Arch-Router, and the preference-aligned work we do in RouterServicev1?

There isn't any change in that. These new fields are optional and are not used with router-request.

salmanap: Okay. approved. Thanks for creating the new issues.

salmanap (approved):

PR #544, opened 2025-07-26
archgw_model_server: use sys.executable for uv tool install compat

kafonek:
If you install `archgw` / `archgw_modelserver` with `uv` (`uv tool install archgw_modelserver` or `uv tool install --editable ./model_server` for local dev), then `python` may not be an executable on the system. `sys.executable` should behave the same way if you're activating a venv, but also play nicely with `uv tool` workflows


adilhafeez:
This is a harmless change and I am going to merge this as I have manually tested it locally.

PR #547, opened 2025-08-06
pushing new apis module for hermes

salmanap:


crates/hermesllm/Cargo.toml
```diff
@@ -8,3 +8,7 @@ serde = {version = "1.0.219", features = ["derive"]}
 serde_json = "1.0.140"
 serde_with = "3.12.0"
 thiserror = "2.0.12"
+
+[dev-dependencies]
+async-openai = "0.29.0"
+tokio = { version = "1.0", features = ["full"] }
```
adilhafeez: you don't need this

crates/hermesllm/src/apis/openai.rs
```diff
@@ -0,0 +1,1386 @@
...
+            OpenAIApi::ChatCompletions => "/v1/chat/completions",
+        }
+    }
+
+    fn from_endpoint(endpoint: &str) -> Option<Self> {
+        match endpoint {
+            "/v1/chat/completions" => Some(OpenAIApi::ChatCompletions),
+            _ => None,
+        }
+    }
```
adilhafeez, line 34: this alone may not be enough - for example vllm and openai could share same endpoint (/v1/chat/completions)

crates/hermesllm/src/apis/openai.rs
```diff
@@ -0,0 +1,1386 @@
...
+    /// The contents of the message (can be null for some cases)
+    pub content: Option<String>,
+    /// The refusal message generated by the model
+    pub refusal: Option<String>,
+    /// Annotations for the message, when applicable, as when using the web search tool
+    pub annotations: Option<Vec<Value>>,
+    /// If the audio output modality is requested, this object contains data about the audio response
+    pub audio: Option<Value>,
+    /// Deprecated and replaced by tool_calls. The name and arguments of a function that should be called
+    #[serde(skip_serializing_if = "Option::is_none")]
```
adilhafeez: this is not needed

crates/hermesllm/src/apis/openai.rs
```diff
@@ -0,0 +1,1386 @@
...
+}
+
+
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ResponseMessage {
+    pub role: Role,
+    /// The contents of the message (can be null for some cases)
+    pub content: Option<String>,
```
adilhafeez, line 134: could this be non-text response?

crates/hermesllm/src/apis/openai.rs
```diff
@@ -0,0 +1,1386 @@
...
+}
+
+/// Individual content part within a message (text or image)
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(tag = "type")]
+pub enum ContentPart {
+    #[serde(rename = "text")]
+    Text { text: String },
+    #[serde(rename = "image_url")]
+    ImageUrl { image_url: ImageUrl },
```
adilhafeez, line 180: can there be more types? Or those are to be handled later? Like audio?

crates/hermesllm/src/apis/openai.rs
```diff
@@ -0,0 +1,1386 @@
...
+    pub description: Option<String>,
+    pub parameters: Value,
+    pub strict: Option<bool>,
+}
+
+/// Tool choice configuration
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum ToolChoice {
+    String(String), // "none", "auto", "required"
```
adilhafeez: if you know what possible values are then maybe use enum?

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -0,0 +1,130 @@
...
+        endpoints.push(api.endpoint());
+    }
+
+    // Add all Anthropic endpoints
+    for api in AnthropicApi::all_variants() {
+        endpoints.push(api.endpoint());
+    }
+
+    endpoints
+}
```
adilhafeez, line 55: as pointed earlier this may not work as multiple providers could share same api path

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -0,0 +1,130 @@
...
+    if OpenAIApi::from_endpoint(endpoint).is_some() {
+        return Some("openai");
+    }
+
+    if AnthropicApi::from_endpoint(endpoint).is_some() {
+        return Some("anthropic");
+    }
+
+    None
+}
```
adilhafeez, line 68: Will not work - see comment above. Maybe remove this functionality for now.

crates/hermesllm/src/apis/anthropic.rs
```diff
@@ -0,0 +1,898 @@
...
+        delta: MessagesMessageDelta,
+        usage: MessagesUsage,
+    },
+    MessageStop,
+    Ping,
+}
+
+#[skip_serializing_none]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct MessagesStreamMessage {
```
adilhafeez, line 337: MessagesStreamResponse?

crates/hermesllm/src/clients/transformer.rs
```diff
@@ -0,0 +1,1713 @@
...
+
+        // Convert system prompt to system message if present
+        if let Some(system) = req.system {
+            openai_messages.push(system.into());
+        }
+
+        // Convert messages
+        for message in req.messages {
+            let converted_messages: Vec<Message> = message.try_into()?;
+            openai_messages.extend(converted_messages);
```
adilhafeez, line 95: why .extend? why not .push?

adilhafeez: oh wait so message from anthropic can generate more than one message for openai?

crates/hermesllm/src/clients/transformer.rs
```diff
@@ -0,0 +1,1713 @@
...
+            n: None,
+            seed: None,
+            response_format: None,
+            service_tier: None,
+            store: None,
+            metadata: None,
+            modalities: None,
+            function_call: None,
+            functions: None,
+            prediction: None,
```
adilhafeez: use Default to fill these (e.g. `..Default::default()`)

crates/hermesllm/src/clients/transformer.rs
```diff
@@ -0,0 +1,1713 @@
...
+trait ContentUtils<T> {
+    fn extract_tool_calls(&self) -> Result<Option<Vec<ToolCall>>, TransformError>;
+    fn split_for_openai(&self) -> Result<(Vec<ContentPart>, Vec<ToolCall>, Vec<(String, String, bool)>), TransformError>;
+}
+
+// ============================================================================
+// MAIN REQUEST TRANSFORMATIONS
+// ============================================================================
+
+impl TryFrom<MessagesRequest> for ChatCompletionsRequest {
```
adilhafeez: MessagesRequest seem generic in this context. You can rename it so its more clear in this file for example you could do sth like this at top of the file,

```
type AnthropicMessagesRequest = MessagesRequest
```

crates/hermesllm/src/clients/transformer.rs
```diff
@@ -0,0 +1,1713 @@
...
+
+        // Convert tools and tool choice
+        let anthropic_tools = req.tools.map(|tools| convert_openai_tools(tools));
+        let anthropic_tool_choice = convert_openai_tool_choice(req.tool_choice, req.parallel_tool_calls);
+
+        Ok(MessagesRequest {
+            model: req.model,
+            system: system_prompt,
+            messages,
+            max_tokens: req.max_tokens.unwrap_or(4096),
```
adilhafeez: use const for 4096

crates/hermesllm/src/clients/transformer.rs
```diff
@@ -0,0 +1,1713 @@
...
+        let finish_reason: FinishReason = resp.stop_reason.into();
+        let tool_calls = resp.content.extract_tool_calls()?;
+
+        // Convert MessageContent to String for response
+        let content_string = match content {
+            MessageContent::Text(text) => Some(text),
+            MessageContent::Parts(parts) => {
+                let text = parts.extract_text();
+                if text.is_empty() { None } else { Some(text) }
+            }
```
adilhafeez, line 180: what if parts contained image_url, here you are forcing it to be text which isn't correct. If image type is specified you wan to use image_url type.

crates/hermesllm/Cargo.toml
salmanap: Fixed

crates/hermesllm/src/apis/openai.rs
salmanap: well however sends the /v1/chat/completions call to us is stating that the API they are calling is one from OpenAI. So this would work from any client that is sending v1/chat/completions

crates/hermesllm/src/apis/openai.rs
salmanap: According to https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages right now the chat/completions can take only two inputs text or imageUrl

crates/hermesllm/src/clients/endpoints.rs
salmanap: that wouldn't matter to us - if they are using the /v1/messages or /v1/chat/completions API our default will be to match it to the OpenAI or Anthropic objects

crates/hermesllm/src/apis/anthropic.rs
salmanap: Agree - let me fix

crates/hermesllm/src/clients/transformer.rs
salmanap: system is a top level property in anthropic. in OpenAI the same property is pushed in the messages[] object under role=="system", content="..."

crates/hermesllm/src/clients/endpoints.rs
salmanap: should work - all we want to do is to detect how to unmarshall the request object and the clients we support are either OpenAI or Anthropic compliant.

crates/hermesllm/src/clients/transformer.rs
salmanap: Fixed.

crates/hermesllm/src/clients/transformer.rs
salmanap: Fixed.

crates/hermesllm/src/apis/openai.rs
salmanap: Fixed.

crates/hermesllm/src/clients/transformer.rs
salmanap: fixed.

crates/hermesllm/src/clients/transformer.rs
salmanap: This is for response - the response will always have text in v1/messages. If we were to add support for v1/images that would be a different story

crates/hermesllm/src/apis/openai.rs
salmanap: Fair. fixed

crates/hermesllm/src/apis/openai.rs
adilhafeez: We discussed this in person for now we will assume that all requests coming in with request path of `/v1/chat/completions` will default to openai

adilhafeez (commented):
In rust from/into are interchangeable - https://doc.rust-lang.org/rust-by-example/conversion/from_into.html#from-and-into-are-interchangeable

So you only need to implement From/TryFrom and into will work as compiler can check the type at compile time.

adilhafeez (commented):
For transformer I think we should use following directory structure,

/providers/
- anthropic.rs
- openai.rs
- gemini.rs

/adapters
- anthropic.rs
- openai.rs
- gemini.rs

Here adapters/openai.rs will contain translation logic for converting openai messages to anthropic and to gemini

And for adapters/gemini.rs will contain log for converting gemini request/response to openai and to anthropic.

adilhafeez (approved):
> For transformer I think we should use following directory structure,
> 
> /providers/
> 
> anthropic.rs
> openai.rs
> gemini.rs
> /adapters
> 
> anthropic.rs
> openai.rs
> gemini.rs
> Here adapters/openai.rs will contain translation logic for converting openai > messages to anthropic and to gemini
> 
> And for adapters/gemini.rs will contain log for converting gemini request/response to openai and to anthropic.



sync'ed offline we will pick these transformation once we add support for other providers like gemini

PR #551, opened 2025-08-11
fix cve_2025-6020 by removing libpam

adilhafeez:
Fixes cve by removing libpam (passwd, login, adduser won't work but we don't use them in container anyway so we should be good)


salmanap (approved):

salmanap (commented):
Please review CoPilots' requested edits

PR #558, opened 2025-08-22
add support for v1/messages and transformations

salmanap:
Adding support for Anthropic v1/messages API so that we can support Claude code

### DIFF
--- a/crates/brightstaff/src/handlers/chat_completions.rs
+++ b/crates/brightstaff/src/handlers/chat_completions.rs
@@ -5,7 +5,7 @@ use common::configuration::ModelUsagePreference;
 use common::consts::ARCH_PROVIDER_HINT_HEADER;
 use hermesllm::apis::openai::ChatCompletionsRequest;
 use hermesllm::clients::SupportedAPIs;
-use hermesllm::ProviderRequestType;
+use hermesllm::{ProviderRequest, ProviderRequestType};
 use http_body_util::combinators::BoxBody;
 use http_body_util::{BodyExt, Full, StreamBody};
 use hyper::body::Frame;
@@ -29,13 +29,13 @@ pub async fn chat(
     router_service: Arc<RouterService>,
     full_qualified_llm_provider_url: String,
 ) -> Result<Response<BoxBody<Bytes, hyper::Error>>, hyper::Error> {
+
     let request_path = request.uri().path().to_string();
     let mut request_headers = request.headers().clone();
-
     let chat_request_bytes = request.collect().await?.to_bytes();
 
     debug!("Received request body (raw utf8): {}", String::from_utf8_lossy(&chat_request_bytes));
-    let provider_request = match ProviderRequestType::try_from((&chat_request_bytes[..], &SupportedAPIs::from_endpoint(request_path.as_str()).unwrap())) {
+    let mut client_request = match ProviderRequestType::try_from((&chat_request_bytes[..], &SupportedAPIs::from_endpoint(request_path.as_str()).unwrap())) {
         Ok(request) => request,
         Err(err) => {
             warn!("Failed to parse request as ProviderRequestType: {}", err);
@@ -46,9 +46,18 @@ pub async fn chat(
         }
     };
 
-    // Convert to ChatCompletionsRequest regardless of input type
+    // Clone metadata for routing and remove archgw_preference_config from original
+    let routing_metadata = client_request.metadata().clone();
+
+    if client_request.remove_metadata_key("archgw_preference_config") {
+        debug!("Removed archgw_preference_config from metadata");
+    }
+
+    let client_request_bytes_for_upstream = ProviderRequestType::to_bytes(&client_request).unwrap();
+
+    // Convert to ChatCompletionsRequest regardless of input type (clone to avoid moving original)
     let chat_completions_request_for_arch_router: ChatCompletionsRequest =
-        match ProviderRequestType::try_from((provider_request, &SupportedAPIs::OpenAIChatCompletions(hermesllm::apis::OpenAIApi::ChatCompletions))) {
+        match ProviderRequestType::try_from((client_request, &SupportedAPIs::OpenAIChatCompletions(hermesllm::apis::OpenAIApi::ChatCompletions))) {
             Ok(ProviderRequestType::ChatCompletionsRequest(req)) => req,
             Ok(ProviderRequestType::MessagesRequest(_)) => {
                 // This should not happen after conversion to OpenAI format
@@ -67,7 +76,6 @@ pub async fn chat(
             }
         };
 
-
     debug!(
         "[BRIGHTSTAFF -> ARCH_ROUTER] REQ: {}",
         &serde_json::to_string(&chat_completions_request_for_arch_router).unwrap()
@@ -79,7 +87,7 @@ pub async fn chat(
         .map(|(_, value)| value.to_str().unwrap_or_default().to_string());
 
     let usage_preferences_str: Option<String> =
-        chat_completions_request_for_arch_router.metadata.as_ref().and_then(|metadata| {
+        routing_metadata.as_ref().and_then(|metadata| {
             metadata
                 .get("archgw_preference_config")
                 .map(|value| value.to_string())
@@ -155,30 +163,13 @@ pub async fn chat(
             header::HeaderValue::from_str(&trace_parent).unwrap(),
         );
     }
-
-        // remove metadata from the request for downstream calls
-    let mut chat_request_user_preferences_removed = chat_completions_request_for_arch_router.clone();
-    if let Some(ref mut metadata) = chat_request_user_preferences_removed.metadata {
-        metadata.remove("archgw_preference_config");
-        debug!("Removed archgw_preference_config from metadata");
-
-        // if metadata is empty, remove it
-        if metadata.is_empty() {
-            chat_request_user_preferences_removed.metadata = None;
-            debug!("Removed empty metadata from request");
-        }
-    }
-
-    let chat_request_parsed_bytes =
-        serde_json::to_string(&chat_request_user_preferences_removed).unwrap();
-
     // remove content-length header if it exists
     request_headers.remove(header::CONTENT_LENGTH);
 
     let llm_response = match reqwest::Client::new()
         .post(full_qualified_llm_provider_url)
         .headers(request_headers)
-        .body(chat_request_parsed_bytes)
+        .body(client_request_bytes_for_upstream)
         .send()
         .await
     {
--- a/crates/hermesllm/src/apis/anthropic.rs
+++ b/crates/hermesllm/src/apis/anthropic.rs
@@ -497,6 +497,18 @@ impl ProviderRequest for MessagesRequest {
             source: Some(Box::new(e)),
         })
     }
+
+    fn metadata(&self) -> &Option<HashMap<String, Value>> {
+       return  &self.metadata;
+    }
+
+    fn remove_metadata_key(&mut self, key: &str) -> bool {
+        if let Some(ref mut metadata) = self.metadata {
+            metadata.remove(key).is_some()
+        } else {
+            false
+        }
+    }
 }
 
 impl MessagesResponse {
--- a/crates/hermesllm/src/apis/openai.rs
+++ b/crates/hermesllm/src/apis/openai.rs
@@ -81,7 +81,7 @@ pub struct ChatCompletionsRequest {
     // Maximum tokens in the response has been deprecated, but we keep it for compatibility
     pub max_tokens: Option<u32>,
     pub modalities: Option<Vec<String>>,
-    pub metadata: Option<HashMap<String, String>>,
+    pub metadata: Option<HashMap<String, Value>>,
     pub n: Option<u32>,
     pub presence_penalty: Option<f32>,
     pub parallel_tool_calls: Option<bool>,
@@ -598,6 +598,18 @@ impl ProviderRequest for ChatCompletionsRequest {
             message: format!("Failed to serialize OpenAI request: {}", e),
             source: Some(Box::new(e)),
         })
+    }
+
+    fn metadata(&self) -> &Option<HashMap<String, Value>> {
+        return &self.metadata;
+    }
+
+    fn remove_metadata_key(&mut self, key: &str) -> bool {
+        if let Some(ref mut metadata) = self.metadata {
+            metadata.remove(key).is_some()
+        } else {
+            false
+        }
     }
 }
 
--- a/crates/hermesllm/src/providers/request.rs
+++ b/crates/hermesllm/src/providers/request.rs
@@ -1,8 +1,12 @@
 use crate::apis::openai::ChatCompletionsRequest;
 use crate::apis::anthropic::MessagesRequest;
 use crate::clients::endpoints::SupportedAPIs;
+
+use serde_json::Value;
 use std::error::Error;
 use std::fmt;
+use std::collections::HashMap;
+#[derive(Clone)]
 pub enum ProviderRequestType {
     ChatCompletionsRequest(ChatCompletionsRequest),
     MessagesRequest(MessagesRequest),
@@ -26,6 +30,11 @@ pub trait ProviderRequest: Send + Sync {
 
     /// Convert the request to bytes for transmission
     fn to_bytes(&self) -> Result<Vec<u8>, ProviderRequestError>;
+
+    fn metadata(&self) -> &Option<HashMap<String, Value>>;
+
+    /// Remove a metadata key from the request and return true if the key was present
+    fn remove_metadata_key(&mut self, key: &str) -> bool;
 }
 
 impl ProviderRequest for ProviderRequestType {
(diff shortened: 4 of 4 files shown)


arch/supervisord.conf
```diff
@@ -9,7 +9,7 @@ stdout_logfile_maxbytes=0
 stderr_logfile_maxbytes=0
 
 [program:envoy]
-command=/bin/sh -c "python /app/config_generator.py && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:debug 2>&1 | tee /var/log//envoy.log"
+command=/bin/sh -c "python /app/config_generator.py && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:debug --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log//envoy.log"
```
adilhafeez, line 12: let's take it up in a separate code change

tests/e2e/run_e2e_tests.sh
```diff
@@ -48,6 +48,8 @@ cd ../../
 archgw build
 cd -
 
+poetry install
```
adilhafeez, line 52: maybe add a comment here why we're doing poetry install here, also think we should maybe have a separate env just for e2e tests. Let's pick that up in next change.

tests/e2e/test_prompt_gateway.py
```diff
@@ -352,3 +355,178 @@ def test_prompt_gateway_prompt_guard_jailbreak(stream):
             response_json.get("choices")[0]["message"]["content"]
             == "Looks like you're curious about my abilities, but I can only provide assistance for weather forecasting."
         )
+
+
+def test_claude_v1_messages_api():
+    """Test Claude client using /v1/messages API through llm_gateway (port 12000)"""
+    # Get the base URL from the LLM gateway endpoint
+    base_url = LLM_GATEWAY_ENDPOINT.replace("/v1/chat/completions", "")
```
adilhafeez, line 363: so why not use http://localhost:12000/v1/messages?

tests/e2e/test_prompt_gateway.py
```diff
@@ -352,3 +355,178 @@ def test_prompt_gateway_prompt_guard_jailbreak(stream):
...
+                "content": "Hello, please respond with exactly: Hello from Claude!",
+            }
+        ],
+    )
+
+    assert message.content[0].text == "Hello from Claude!"
+
+
+def test_claude_v1_messages_api_streaming():
+    base_url = LLM_GATEWAY_ENDPOINT.replace("/v1/chat/completions", "")
```
adilhafeez, line 384: same comment as above - why not use http://localhost:12000/v1/messages endpoint?

tests/e2e/test_prompt_gateway.py
```diff
@@ -352,3 +355,178 @@ def test_prompt_gateway_prompt_guard_jailbreak(stream):
...
+        ],
+    ) as stream:
+        # This yields only text deltas in order
+        pieces = [t for t in stream.text_stream]
+        full_text = "".join(pieces)
+
+        # You can also get the fully-assembled Message object
+        final = stream.get_final_message()
+        # A safe way to reassemble text from the content blocks:
+        final_text = "".join(b.text for b in final.content if b.type == "text")
```
adilhafeez, line 405: you're computing full_text already at like 400 why compute it again?

```
        full_text = "".join(pieces)
```

adilhafeez: basically all tests below have same issue with computing full text and the endpoint url

crates/llm_gateway/src/stream_context.rs
```diff
@@ -227,10 +593,35 @@ impl HttpContext for StreamContext {
...
             self.select_llm_provider();
+
+            // Check if this is a supported API endpoint
+            if SupportedAPIs::from_endpoint(&request_path).is_none() {
+                self.send_http_response(404, vec![], Some(b"Unsupported endpoint"));
+                return Action::Continue;
+            }
+
+            // Get the SupportedApi for routing decisions
+            let supported_api: Option<SupportedAPIs> = SupportedAPIs::from_endpoint(&request_path);
```
adilhafeez, line 606: This is over simplifying the matching. It may not work for custom providers. I think adding host name (for example api.anthropic.com or api.openai.com etc.) would make it clear and less error prone. And will open us to add custom providers.

crates/llm_gateway/src/stream_context.rs
```diff
@@ -227,10 +593,35 @@ impl HttpContext for StreamContext {
...
+            self.client_api = supported_api;
```
adilhafeez, line 607: why is supported_apis called client_api?

crates/llm_gateway/src/stream_context.rs
```diff
@@ -227,10 +593,35 @@ impl HttpContext for StreamContext {
...
+
+            // Debug: log provider, client API, resolved API, and request path
+            if let (Some(api), Some(provider)) =
+                (self.client_api.as_ref(), self.llm_provider.as_ref())
+            {
+                let provider_id = provider.to_provider_id();
+                self.resolved_api = Some(provider_id.compatible_api_for_client(api));
+            } else {
+                self.resolved_api = None;
+            }
```
adilhafeez, line 617: provider/model name could still change based on what we see in message body so resolved_api would need to wait until on_request_body is complete

crates/llm_gateway/src/stream_context.rs
```diff
@@ -83,6 +97,18 @@ impl StreamContext {
...
+    fn update_upstream_path(&mut self, request_path: &str) {
+        let hermes_provider_id = self.llm_provider().to_provider_id();
+        if let Some(api) = &self.client_api {
+            let target_endpoint =
+                api.target_endpoint_for_provider(&hermes_provider_id, request_path);
+            if target_endpoint != request_path {
+                self.set_http_request_header(":path", Some(&target_endpoint));
+            }
+        }
+    }
```
adilhafeez, line 107: we need to wait for on_request_body to decide the model name before we make path changes

crates/llm_gateway/src/stream_context.rs
```diff
@@ -135,9 +140,23 @@ impl StreamContext {
...
 
-        let authorization_header_value = format!("Bearer {}", llm_provider_api_key_value);
-
-        self.set_http_request_header("Authorization", Some(&authorization_header_value));
+        // Set API-specific headers based on the resolved upstream API
+        match self.resolved_api.as_ref() {
+            Some(SupportedAPIs::AnthropicMessagesAPI(_)) => {
+                // Anthropic API requires x-api-key and anthropic-version headers
+                // Remove any existing Authorization header since Anthropic doesn't use it
+                self.set_http_request_header("Authorization", None);
```
adilhafeez: should use `remove_http_request_header`

crates/llm_gateway/src/stream_context.rs
```diff
@@ -135,9 +140,23 @@ impl StreamContext {
...
+                self.set_http_request_header("x-api-key", Some(llm_provider_api_key_value));
+                self.set_http_request_header("anthropic-version", Some("2023-06-01"));
+            }
+            Some(SupportedAPIs::OpenAIChatCompletions(_)) | None => {
+                // OpenAI and default: use Authorization Bearer token
+                // Remove any existing x-api-key header since OpenAI doesn't use it
+                self.set_http_request_header("x-api-key", None);
```
adilhafeez: use `remove_http_request_header`

crates/llm_gateway/src/stream_context.rs
```diff
@@ -73,6 +77,16 @@ impl StreamContext {
...
+    /// Returns the appropriate request identifier for logging.
+    /// Uses request_id (from x-request-id header) when available, otherwise falls back to context_id.
+    fn request_identifier(&self) -> String {
+        self.request_id
+            .as_ref()
+            .filter(|id| !id.is_empty()) // Filter out empty strings
+            .map(|id| id.clone())
+            .unwrap_or_else(|| self.context_id.to_string())
+    }
     fn llm_provider(&self) -> &LlmProvider {
```
adilhafeez: so context_id is not same as request_id. context_id is internal to envoy and is used to identify requests within envoy

crates/llm_gateway/src/stream_context.rs
```diff
@@ -179,26 +198,375 @@ impl StreamContext {
...
+
+    fn debug_log_body(&self, body: &[u8]) {
+        debug!(
+            "[ARCHGW_REQ_ID:{}] UPSTREAM_RAW_RESPONSE: body_size={} content={}",
+            self.request_identifier(),
+            body.len(),
+            String::from_utf8_lossy(body)
+        );
+    }
+
```
adilhafeez, line 397: this is unnecessary

crates/llm_gateway/src/stream_context.rs
salmanap: all requests go through brightstaff via port 12000 - it resolves the provider_hint and model. So this shouldn't be necessary.

arch/supervisord.conf
salmanap: Its really hard to follow the logs especially with this transformation without some structure. Its a small change, I would advocate strongly that we clean up our logs so that we incrementally make it easier for developers to find things. For, example, Dennis had a tough time navigating our Docker logs.

crates/llm_gateway/src/stream_context.rs
salmanap: Why? The provider-hint header from brightstaff resolves the provider. Plus the client API isn't mutated. The combination of those two things resolve the upstream API neatly. No?

crates/llm_gateway/src/stream_context.rs
```diff
@@ -179,26 +198,375 @@ impl StreamContext {
...
+        debug!(
+            "[ARCHGW_REQ_ID:{}] STREAMING_PROCESS: provider_id={:?} chunk_size={}",
+            self.request_identifier(),
+            provider_id,
+            body.len()
+        );
+        match self.client_api.as_ref() {
+            Some(client_api) => {
+                let client_api = client_api.clone(); // Clone to avoid borrowing issues
+                let upstream_api = provider_id.compatible_api_for_client(&client_api);
```
adilhafeez, line 412: where is this used in the code?

crates/llm_gateway/src/stream_context.rs
```diff
@@ -179,26 +198,375 @@ impl StreamContext {
...
+                            Err(e) => {
+                                warn!(
+                                    "[ARCHGW_REQ_ID:{}] STREAMING_CHUNK_ERROR: {}",
+                                    self.request_identifier(),
+                                    e
+                                );
+                                return Err(Action::Continue);
+                            }
+                        }
+                    }
```
adilhafeez, line 473: this could be extracted away in separate function so keep code simpler

crates/common/src/routing.rs
```diff
@@ -33,6 +33,8 @@ pub fn get_llm_provider(
         return provider;
     }
 
+    //This is a fallback to the default provider if no specific provider is found.
+    //For example, if the client sends in gpt-4-1 and that's not configured in arch_config, we fall back to the default.
     if llm_providers.default().is_some() {
```
adilhafeez: not needed

crates/hermesllm/src/apis/anthropic.rs
```diff
@@ -369,6 +396,109 @@ impl MessagesRequest {
...
+    fn prompt_tokens(&self) -> usize {
+        self.usage.input_tokens as usize
+    }
+    fn total_tokens(&self) -> usize {
+        (self.usage.input_tokens + self.usage.output_tokens) as usize
+    }
+}
+
+impl MessagesResponse {
+    pub fn usage(&self) -> Option<&dyn TokenUsage> {
```
adilhafeez: ... impl ProviderResponse for MessagesResponse

crates/hermesllm/src/providers/response.rs
```diff
@@ -1,105 +1,379 @@
...
-    fn content_delta(&self) -> Option<&str>;
+// TryFrom implementation to convert raw bytes to SseEvent with parsed provider response
+impl TryFrom<(&SseEvent, &SupportedAPIs, &SupportedAPIs)> for SseEvent {
+    type Error = Box<dyn std::error::Error + Send + Sync>;
 
-    /// Check if this is the final chunk in the stream
-    fn is_final(&self) -> bool;
+    fn try_from((sse_event, client_api, upstream_api): (&SseEvent, &SupportedAPIs, &SupportedAPIs)) -> Result<Self, Self::Error> {
+        // Create a new transformed event based on the original
+        let mut transformed_event = sse_event.clone();
```
adilhafeez: consume it directly there

crates/hermesllm/src/providers/response.rs
```diff
@@ -1,105 +1,379 @@
...
+    pub data: Option<String>,  // The JSON payload after "data: "
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub event: Option<String>,  // Optional event type (e.g., "message_start", "content_block_delta")
+
+    #[serde(skip_serializing, skip_deserializing)]
+    pub raw_line: String,  // The complete line as received including "data: " prefix and "\n\n"
+
+     #[serde(skip_serializing, skip_deserializing)]
+    pub raw_line_transformed: String,  // The complete line as received including "data: " prefix and "\n\n"
```
adilhafeez: consider renaming to line_buffer or transformed_buffer - and remove all json serialization primitives

tests/e2e/run_e2e_tests.sh
salmanap: fixed.

tests/e2e/test_prompt_gateway.py
salmanap: The default constant is used to make sure we have one for any type of call, including curl. For using the clients we have to remove that default path and have the client library construct its own path

tests/e2e/test_prompt_gateway.py
salmanap: we should use the client. Not curl to 12000. This where we know that the client can process transformed messages from our proxy appropriately.

crates/llm_gateway/src/stream_context.rs
salmanap: this is in the stream_context.rs file. We are tyring to see what api did the client send to the LLM_gateway

crates/llm_gateway/src/stream_context.rs
salmanap: fixed.

crates/llm_gateway/src/stream_context.rs
salmanap: fixed

crates/llm_gateway/src/stream_context.rs
salmanap: Agree. For now we will return back something like this

   fn request_identifier(&self) -> String {
        self.request_id
            .as_ref()
            .filter(|id| !id.is_empty()) // Filter out empty strings
            .map(|id| id.clone())
            .unwrap_or_else(|| "NO_REQUEST_ID".to_string())
    }

crates/llm_gateway/src/stream_context.rs
salmanap: on line 430

                        match SseEvent::try_from((&sse_event, &client_api, &upstream_api)) {
                            Ok(event) => event,
                            Err(e) => {
                                warn!("Failed to transform SSE event: {}", e);
                                return Err(Action::Continue);
                            }
                        };

crates/llm_gateway/src/stream_context.rs
salmanap: sure. would it be okay if we punt that for a different refactor? As is there are a lot of changes to land

crates/common/src/routing.rs
salmanap: fixed.

crates/hermesllm/src/providers/response.rs
salmanap: fixed.

crates/hermesllm/src/apis/anthropic.rs
salmanap: fixed.

crates/hermesllm/src/providers/response.rs
salmanap: fixed.

crates/brightstaff/src/handlers/chat_completions.rs
```diff
@@ -33,54 +35,42 @@ pub async fn chat(
...
+            let err_msg = format!("Failed to parse request: {}", err);
+            let mut bad_request = Response::new(full(err_msg));
+            *bad_request.status_mut() = StatusCode::BAD_REQUEST;
+            return Ok(bad_request);
         }
+    };
+
+    // Convert to ChatCompletionsRequest regardless of input type
+    let chat_completions_request_for_arch_router: ChatCompletionsRequest =
+        match ProviderRequestType::try_from((provider_request, &SupportedAPIs::OpenAIChatCompletions(hermesllm::apis::OpenAIApi::ChatCompletions))) {
```
adilhafeez: With this change you're effectively removing support for anthropic. Because now if someone sends an anthropic request you will fail right here.

adilhafeez (approved):
pls create an issue in for bright staff to not do any conversion if routing preferences isn't in use

PR #562, opened 2025-09-04
Optionally use different host internal and host ip for networking in dev environments.

dkennetzoracle:
Allows users to optionally build dev environment with non-default HOST_INTERNAL and HOST_IP. This can resolve issues when there are connectivity errors from archgw to the host machine.


arch/docker-compose.dev.yaml
```diff
@@ -8,18 +8,20 @@ services:
       - "12000:12000"
       - "19901:9901"
     volumes:
-      - ${ARCH_CONFIG_FILE:-../demos/samples_python/weather_forecast/arch_config.yaml}:/app/arch_config.yaml
+      - ${ARCH_CONFIG_FILE:-/Users/dkennetz/arch_gateway/arch_config.yaml}:/app/arch_config.yaml
```
dkennetzoracle: missed this one - will fix.

arch/README.md
```diff
@@ -6,10 +6,19 @@
 $ rustup target add wasm32-wasip1
 ```
 
-## Building
+## Container Runtime Support
 
 ```sh
-$ cargo build --target wasm32-wasip1 --release
+# Build the gateway WASM modules (must be run from the crates directory)
```
dkennetzoracle, line 12: Can revert this, but `cargo build --target wasm32-wasip1 --release` this failed on my machine, but building these packages individually worked. I can revert this, I just wanted to draw attention to it.

adilhafeez: That is because llm_gateway and prompt_gateway binaries are loaded in envoy which only supports [wasm32-wasip1](https://doc.rust-lang.org/rustc/platform-support/wasm32-wasip1.html#wasm32-wasip1) runtime. But for brightstaff  there is no such limitation as runs as individual server.

adilhafeez:
@dkennetzoracle thanks a lot for putting this PR together. I think we can also do following too as it will be a bit simpler. What do you think?

```
    extra_hosts:
      - "host.docker.internal:host-gateway"
      - "host.rancher-docker.internal:host-gateway"
```

dkennetzoracle:
> @dkennetzoracle thanks a lot for putting this PR together. I think we can also do following too as it will be a bit simpler. What do you think?
> 
> ```
>     extra_hosts:
>       - "host.docker.internal:host-gateway"
>       - "host.rancher-docker.internal:host-gateway"
> ```

No problem! Happy to help where I can. I ended up finding out that host.docker.internal worked for me, the docker bridge just pointed to the wrong IP - or an IP that failed to connect out.

Summary: either host worked if I set the correct HOST_IP (which was that 192 value). I'm okay either way.

My CLI looks like:
```
ARCH_CONFIG_FILE=/Users/dkennetz/code/arch_demos/oci_apis/arch_config.yaml OPENAI_API_KEY=test MISTRAL_API_KEY=test HOST_INTERNAL=host.docker.internal HOST_IP=192.X.X.X docker-compose -f docker-compose.dev.yaml up archgw
```

PR #564, opened 2025-09-10
Adil/agent format

adilhafeez:


arch/arch_config_schema.yaml
```diff
@@ -8,46 +8,76 @@ properties:
       - v0.1.0
       - 0.1-beta
       - 0.2.0
+      - v0.3.0
+
+  agents:
+    type: array
+    items:
+      type: object
```
salmanap, line 16: Do you want to add the "properties" object. Just like you have for the listeners?

arch/arch_config_schema.yaml
```diff
@@ -8,46 +8,76 @@ properties:
...
+            port:
+              type: integer
+            address:
+              type: string
+            timeout:
+              type: string
+            type:
+              type: string
+              enum:
+                - model_listener
```
salmanap: I would simplify this to "agent" "model" and "prompt"

arch/arch_config_schema.yaml
```diff
@@ -60,7 +90,52 @@ properties:
...
+          type: string
+        default:
+          type: boolean
+        base_url:
+          type: string
+        http_host:
+          type: string
+        provider_interface:
+          type: string
+          enum:
```
salmanap, line 113: Please borrow from all the providers we now support in a first class fashion. That list is available in config_generator.py. We'll need a separate PR to make this more consistent and central

arch/arch_config_schema.yaml
```diff
@@ -282,4 +357,4 @@ properties:
 additionalProperties: false
 required:
   - version
-  - llm_providers
+  - listeners
```
salmanap, line 360: don't we need model_providers || llm_providers as a required property?

arch/envoy.template.yaml
```diff
@@ -218,7 +218,7 @@ static_resources:
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
 
-    - name: egress_api_traffic
+    - name: outbound_api_traffic
```
salmanap, line 221: can we make this more dynamic? Meaning can it be the name of the agent itself or the api/tool we are calling. Else this would show up in the traces/logs as "outbound_api_traffic" which would feel very generic.

arch/envoy.template.yaml
```diff
@@ -292,6 +292,108 @@ static_resources:
...
+                      "@type": type.googleapis.com/envoy.config.trace.v3.OpenTelemetryConfig
+                      grpc_service:
+                        envoy_grpc:
+                          cluster_name: opentelemetry_collector
+                        timeout: 0.250s
+                      service_name: arch_gateway
+                  random_sampling:
+                    value: {{ arch_tracing.random_sampling }}
+                {% endif %}
+                stat_prefix: agents_traffic
```
salmanap: How would the service_name and stat_prefix look like in the logs? I would change the service_name to archgw and the stat prefix to ":agent:"

arch/tools/cli/utils.py
```diff
@@ -37,16 +37,100 @@ def has_ingress_listener(arch_config_file):
...
+            prompt_gateway_listener["timeout"] = ingress_traffic.get(
+                "timeout", prompt_gateway_listener["timeout"]
+            )
+            updated_listeners.append(prompt_gateway_listener)
+
+        return updated_listeners, llm_gateway_listener, prompt_gateway_listener
+
+    model_provider_set = False
+    for listener in listeners:
+        if listener.get("type") == "model_listener":
```
salmanap, line 99: The object is listener the type should just be "models"

arch/tools/cli/utils.py
```diff
@@ -37,16 +37,100 @@ def has_ingress_listener(arch_config_file):
...
+    listeners: dict | list, model_providers: list | None
+) -> tuple[list, dict | None, dict | None]:
+    llm_gateway_listener = {
+        "name": "egress_traffic",
+        "type": "model_listener",
+        "port": 12000,
+        "address": "0.0.0.0",
+        "timeout": "30s",
+        "model_providers": model_providers or [],
+        "protocol": "openai",
```
salmanap: I am not sure if we need protocol for models. That's now based on v1/chat/completions v1/responses and /v1/messages. This is essentially now based on the model provider they configure. This is also documented in our docs.

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -0,0 +1,145 @@
...
+use super::response_handler::ResponseHandler;
+use crate::router::llm_router::RouterService;
+
+/// Main errors for agent chat completions
+#[derive(Debug, thiserror::Error)]
+pub enum AgentChatError {
+    #[error("Agent selection error: {0}")]
+    Selection(#[from] AgentSelectionError),
+    #[error("Pipeline processing error: {0}")]
+    Pipeline(#[from] PipelineError),
```
salmanap, line 21: I think we should call it AgentFilterChain. Just easier to map from external concepts to internal ones. Create a naming mismatch will make it slightly hard to read the code and see how it implements the concepts we expose via our yaml config file.

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -0,0 +1,145 @@
...
+    #[error("Response handling error: {0}")]
+    Response(#[from] super::response_handler::ResponseError),
+    #[error("Request parsing error: {0}")]
+    RequestParsing(#[from] serde_json::Error),
+    #[error("HTTP error: {0}")]
+    Http(#[from] hyper::Error),
+}
+
+pub async fn agent_chat(
```
salmanap, line 30: I think we briefly discussed this earlier. Should we make this more generic in the sense that we call it agent_message_handler? this way if the input is simply a string we can handle that too?

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -0,0 +1,145 @@
+use std::sync::Arc;
```
salmanap, line 1: This is such cleanly written and formatted code. Love!

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -0,0 +1,145 @@
...
+    // Parse request body
+    let request_headers = request.headers().clone();
+    let chat_request_bytes = request.collect().await?.to_bytes();
+
+    debug!(
+        "Received request body (raw utf8): {}",
+        String::from_utf8_lossy(&chat_request_bytes)
+    );
+
+    let chat_completions_request: ChatCompletionsRequest =
```
salmanap, line 112: I think for the incoming traffic, we can take two approaches. Support a simple text-based input and convert that to a v1/chat/completions endpoint. Or simply pass that through as is. In that case, the filter_chain will expect a string output response else would default to the messages array.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -0,0 +1,228 @@
...
+    pub async fn process_filter_chain(
+        &self,
+        initial_request: &ChatCompletionsRequest,
+        agent_pipeline: &AgentPipeline,
+        agent_map: &HashMap<String, Agent>,
+        request_headers: &HeaderMap,
+    ) -> Result<Vec<Message>, PipelineError> {
+        let mut chat_completions_history = initial_request.messages.clone();
+
+        for agent_name in &agent_pipeline.filter_chain {
```
salmanap: for agent_filter_name in agent_filter_chain.get

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -0,0 +1,228 @@
...
+                        agent_name, err, response_content
+                    )
+                })?;
+        }
+
+        Ok(chat_completions_history)
+    }
+
+    /// Send request to a specific agent and return the response content
+    async fn send_agent_request(
```
salmanap: send_agent_filter_chain_request

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -0,0 +1,228 @@
...
+            .and_then(|message| message.get("content"))
+            .and_then(|content| content.as_str())
+            .ok_or_else(|| PipelineError::NoContentInResponse(agent.id.clone()))?
+            .to_string();
+
+        Ok(content)
+    }
+
+    /// Send request to terminal agent and return the raw response for streaming
+    pub async fn send_terminal_request(
```
salmanap: shouldn't this be handled now in the agent_chat portion. Meaning we have removed the notion of a termainl_agent from the filter chain. The agent that we detected will get the request after the filter_chain has been processed.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        let response = self
+            .client
+            .post(&self.url)
+            .headers(agent_headers)
+            .body(request_body)
+            .send()
+            .await?;
+
+        let response_bytes = response.bytes().await?;
```
salmanap, line 125: How should we handle filter_chain failures? Silent? Catastrophic? We will need to define the lifecycle of the filter chain and add that to our docs. This is important to get right early on. Can we borrow from Envoy's thinking here?

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -0,0 +1,228 @@
...
+
+        // Parse the response as JSON to extract the content
+        let response_json: serde_json::Value = serde_json::from_slice(&response_bytes)?;
+
+        let content = response_json
```
salmanap, line 130: Why aren't we using ChatCompletionResponse::try_from here?

crates/brightstaff/src/handlers/response_handler.rs
```diff
@@ -0,0 +1,160 @@
+use bytes::Bytes;
```
salmanap, line 1: I think i'll need a bit more context for how you intend to use this and where. It wasn't clear to me what this utility was doing above and beyond what the standard Rust traits would offer in terms of response handling. DISCUSSION

crates/brightstaff/src/handlers/response_handler.rs
```diff
@@ -0,0 +1,160 @@
...
+        Self::create_error_response(StatusCode::BAD_REQUEST, message)
+    }
+
+    /// Create an internal server error response
+    pub fn create_internal_error(message: &str) -> Response<BoxBody<Bytes, hyper::Error>> {
+        Self::create_error_response(StatusCode::INTERNAL_SERVER_ERROR, message)
+    }
+
+    /// Create a streaming response from a reqwest response
+    pub async fn create_streaming_response(
```
salmanap, line 70: Should this be used for the LLM streaming traffic as well? DISCUSSION

arch/arch_config_schema.yaml
adilhafeez: `type: array` allows us to add list while type property is used for dict types. Here is we need agents to be a list.

arch/arch_config_schema.yaml
adilhafeez: that will happen when I rebase with main

arch/arch_config_schema.yaml
adilhafeez: It is possible to have an agent without llm_providers. Same way we can have prompt listener without llm_providers.

arch/envoy.template.yaml
adilhafeez: This is internal listener that we use to route traffic from envoy to other services that are registered. We don't have per service listener. If we did that would bloat listeners section and could take more resources.

arch/envoy.template.yaml
adilhafeez: it should be `listener.name` rather than generic agents_traffic. I think it should be `agents_{{listener.name}}`

arch/tools/cli/utils.py
adilhafeez: I will take a it out - it was there for legacy reasons

arch/tools/cli/utils.py
adilhafeez: yes that was from same feedback from above

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: sure - will update

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: It does so in `handle_agent_chat` function down below. We should be able to handle generic input but for this code change I wanted to focus on chat completion.

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: as discussed earlier - for this change I wanted to focus on chat completions protocol for more generic approach a followup change will be needed

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: I think its the same name change we discussed in the beginning of the PR - will take care of it

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: got it!

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: errors will be propagated appropriately back to the developer. Notice the usage of `Ok(...)` it uses rust's [result module](https://doc.rust-lang.org/std/result/) to handle errors and appropriately maps them to errors.

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: We are not interested in looking inside the message body here we're just streaming results back to developer. For token counting envoy is already handling that part. Here in bright_staff we're just relaying message back to developer for final response.

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: this is good observation - I did ported most code to use new pattern this was left behind. Will update.

crates/brightstaff/src/handlers/response_handler.rs
adilhafeez: it just stores utility functions that are used inside other files like pipeline_processor.rs

crates/brightstaff/src/handlers/response_handler.rs
adilhafeez: this change is just for chat completions and its a pass-through for final response

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: updated

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: updated

arch/arch_config_schema.yaml
salmanap: Technically that's true. Should we throw a warning?

PR #571, opened 2025-09-19
Salmanap/fix docs new providers model alias

salmanap:


docs/source/concepts/llm_providers/model_aliases.rst
```diff
@@ -0,0 +1,254 @@
...
+    model_aliases:
+      arch.reasoning.v1:
+        target: gpt-oss-120b
+        guardrails:
+          max_latency: 5s
+          max_cost_per_request: 0.10
+          block_categories: ["jailbreak", "PII"]
+          content_filters:
+            - type: "profanity"
+            - type: "sensitive_data"
```
adilhafeez, line 181: for max_latency - what happens when latency goes beyond threshold? Would we cancel the request? I think this behavior should be configurable. Default fail close.

PR #581, opened 2025-10-01
stream access logs and improve access log format

adilhafeez:


salmanap (approved):
How do you make WASM + Envoy logs consistent looking.

PR #583, opened 2025-10-02
adding support for Qwen models and fixed issue with passing PATH vari…

salmanap:
…able

### DIFF
--- a/arch/tools/cli/main.py
+++ b/arch/tools/cli/main.py
@@ -205,6 +205,8 @@ def up(file, path, service, foreground):
         "MODEL_SERVER_PORT": os.getenv("MODEL_SERVER_PORT", "51000"),
     }
     env = os.environ.copy()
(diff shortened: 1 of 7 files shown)


arch/tools/cli/main.py
```diff
@@ -205,6 +205,8 @@ def up(file, path, service, foreground):
         "MODEL_SERVER_PORT": os.getenv("MODEL_SERVER_PORT", "51000"),
     }
     env = os.environ.copy()
+    # Remove PATH variable if present
+    env.pop("PATH", None)
```
adilhafeez, line 209: this should be a separate PR

adilhafeez (approved):

arch/tools/cli/main.py
salmanap: agree - just movingly quickly.

PR #588, opened 2025-10-11
Add support for Amazon Bedrock Converse and ConverseStream

salmanap:
… for streaming and binary frames

### DIFF
--- a/arch/supervisord.conf
+++ b/arch/supervisord.conf
@@ -9,7 +9,7 @@ stdout_logfile_maxbytes=0
 stderr_logfile_maxbytes=0
 
 [program:envoy]
-command=/bin/sh -c "python /app/config_generator.py && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:info --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log/envoy.log | while IFS= read -r line; do echo '[envoy_logs] ' \"$line\"; done"
+command=/bin/sh -c "python -m cli.config_generator && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:info --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log/envoy.log | while IFS= read -r line; do echo '[envoy_logs] ' \"$line\"; done"
 stdout_logfile=/dev/stdout
 redirect_stderr=true
 stdout_logfile_maxbytes=0
--- a/arch/tools/cli/config_generator.py
+++ b/arch/tools/cli/config_generator.py
@@ -1,9 +1,11 @@
 import json
 import os
+from cli.utils import convert_legacy_listeners
 from jinja2 import Environment, FileSystemLoader
 import yaml
 from jsonschema import validate
 from urllib.parse import urlparse
+from copy import deepcopy
 
 
 SUPPORTED_PROVIDERS = [
@@ -73,17 +75,58 @@ def validate_and_render_schema():
     _ = yaml.safe_load(arch_config_schema)
     inferred_clusters = {}
 
+    # Convert legacy llm_providers to model_providers
+    if "llm_providers" in config_yaml:
+        if "model_providers" in config_yaml:
+            raise Exception(
+                "Please provide either llm_providers or model_providers, not both. llm_providers is deprecated, please use model_providers instead"
+            )
+        config_yaml["model_providers"] = config_yaml["llm_providers"]
+        del config_yaml["llm_providers"]
+
+    listeners, llm_gateway, prompt_gateway = convert_legacy_listeners(
+        config_yaml.get("listeners"), config_yaml.get("model_providers")
+    )
+
+    config_yaml["listeners"] = listeners
+
     endpoints = config_yaml.get("endpoints", {})
+
+    # Process agents section and convert to endpoints
+    agents = config_yaml.get("agents", [])
+    for agent in agents:
+        agent_id = agent.get("id")
+        agent_endpoint = agent.get("url")
+
+        if agent_id and agent_endpoint:
+            urlparse_result = urlparse(agent_endpoint)
+            if urlparse_result.scheme and urlparse_result.hostname:
+                protocol = urlparse_result.scheme
+
+                port = urlparse_result.port
+                if port is None:
+                    if protocol == "http":
+                        port = 80
+                    else:
+                        port = 443
+
+                endpoints[agent_id] = {
+                    "endpoint": urlparse_result.hostname,
+                    "port": port,
+                    "protocol": protocol,
+                }
 
     # override the inferred clusters with the ones defined in the config
     for name, endpoint_details in endpoints.items():
         inferred_clusters[name] = endpoint_details
-        endpoint = inferred_clusters[name]["endpoint"]
-        protocol = inferred_clusters[name].get("protocol", "http")
-        (
-            inferred_clusters[name]["endpoint"],
-            inferred_clusters[name]["port"],
-        ) = get_endpoint_and_port(endpoint, protocol)
+        # Only call get_endpoint_and_port for manually defined endpoints, not agent-derived ones
+        if "port" not in endpoint_details:
+            endpoint = inferred_clusters[name]["endpoint"]
+            protocol = inferred_clusters[name].get("protocol", "http")
+            (
+                inferred_clusters[name]["endpoint"],
+                inferred_clusters[name]["port"],
+            ) = get_endpoint_and_port(endpoint, protocol)
 
     print("defined clusters from arch_config.yaml: ", json.dumps(inferred_clusters))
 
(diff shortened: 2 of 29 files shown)


arch/supervisord.conf
```diff
@@ -9,7 +9,7 @@ stdout_logfile_maxbytes=0
 stderr_logfile_maxbytes=0
 
 [program:envoy]
-command=/bin/sh -c "python /app/config_generator.py && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:info --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log/envoy.log | while IFS= read -r line; do echo '[envoy_logs] ' \"$line\"; done"
+command=/bin/sh -c "python /app/config_generator.py && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:debug --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log/envoy.log | while IFS= read -r line; do echo '[envoy_logs] ' \"$line\"; done"
```
adilhafeez: you may want to revert it to info before merging

crates/common/src/consts.rs
```diff
@@ -11,6 +11,7 @@ pub const MODEL_SERVER_NAME: &str = "model_server";
 pub const ARCH_ROUTING_HEADER: &str = "x-arch-llm-provider";
 pub const MESSAGES_KEY: &str = "messages";
 pub const ARCH_PROVIDER_HINT_HEADER: &str = "x-arch-llm-provider-hint";
+pub const ARCH_IS_STREAMING_HEADER: &str = "x-archgw-streaming-request";
```
adilhafeez: let's rename it to arch as other headers also use `arch` and not `archgw`

arch/tools/cli/config_generator.py
```diff
@@ -21,6 +21,7 @@
     "moonshotai",
     "zhipu",
     "qwen",
+    "amazon_bedrock",
```
adilhafeez, line 26: should we just name is bedrock instead of amazon_bedrock?

arch/tools/cli/config_generator.py
```diff
@@ -130,7 +131,10 @@ def validate_and_render_schema():
         provider = model_name_tokens[0]
         # Validate azure_openai and ollama provider requires base_url
         if (
-            provider == "azure_openai" or provider == "ollama" or provider == "qwen"
+            provider == "azure_openai"
+            or provider == "ollama"
+            or provider == "qwen"
+            or provider == "amazon_bedrock"
```
adilhafeez: For bedrock you need either region name or base_url. I do suggest add support for region in the config so its more consistent with amazon style

https://docs.aws.amazon.com/general/latest/gr/bedrock.html

crates/hermesllm/src/apis/amazon_bedrock.rs
```diff
@@ -0,0 +1,1139 @@
...
+    pub prompt_variables: Option<HashMap<String, PromptVariableValues>>,
+    /// Request metadata for filtering logs
+    #[serde(rename = "requestMetadata")]
+    pub request_metadata: Option<HashMap<String, String>>,
+    /// Additional custom metadata (for internal use)
+    pub metadata: Option<HashMap<String, Value>>,
+    /// Whether this request should use streaming endpoint (internal field, not serialized)
+    #[serde(skip)]
+    pub stream: bool,
+}
```
adilhafeez, line 106: But stream=true/false message is part of ConverseRequest. When upstream is amazon_bedrock wouldn't you want to pass stream=true/false?

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -62,12 +48,21 @@ impl SupportedAPIs {
...
                 match provider_id {
                     ProviderId::Anthropic => "/v1/messages".to_string(),
+                    ProviderId::AmazonBedrock => {
+                        if request_path.starts_with("/v1/") && !is_streaming {
+                            format!("/model/{}/converse", model_id)
+                        } else if request_path.starts_with("/v1/") && is_streaming {
+                            format!("/model/{}/converse-stream", model_id)
+                        } else {
+                            default_endpoint
+                        }
```
adilhafeez:
```suggestion
                        if request_path.starts_with("/v1/") {
                          if is_streaming {
                              format!("/model/{}/converse-stream", model_id)
                          } else {
                              format!("/model/{}/converse", model_id)
                          }
                        } else {
                            default_endpoint
                        }
```

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -108,6 +103,16 @@ impl SupportedAPIs {
...
                     }
+                    ProviderId::AmazonBedrock => {
+                        if request_path.starts_with("/v1/") && !is_streaming {
+                            format!("/model/{}/converse", model_id)
+                        } else if request_path.starts_with("/v1/") && is_streaming {
+                            format!("/model/{}/converse-stream", model_id)
+                        }
+                        else {
+                            default_endpoint
+                        }
```
adilhafeez:
```suggestion
                        if request_path.starts_with("/v1/") {  
                          if is_streaming {  
                              format!("/model/{}/converse-stream", model_id)  
                          } else {  
                              format!("/model/{}/converse", model_id)  
                          }  
                        }  
                        else {
                            default_endpoint
                        }
```

crates/hermesllm/src/providers/request.rs
```diff
@@ -10,6 +13,8 @@ use std::collections::HashMap;
 pub enum ProviderRequestType {
     ChatCompletionsRequest(ChatCompletionsRequest),
     MessagesRequest(MessagesRequest),
+    BedrockConverse(ConverseRequest),
+    BedrockConverseStream(ConverseStreamRequest),
     //add more request types here
 }
```
adilhafeez, line 19: MessagesRequest -> AnthropicMessagesRequest. We should consider this change as it stands right now reading this its not straightforward for me to understand that MessagesRequest is actually anthropic message.

crates/hermesllm/src/providers/request.rs
```diff
@@ -140,14 +161,59 @@ impl TryFrom<(ProviderRequestType, &SupportedAPIs)> for ProviderRequestType {
...
+
+            // Amazon Bedrock to other APIs conversions
+            (ProviderRequestType::BedrockConverse(_), _) => {
+                todo!("Amazon Bedrock to ChatCompletionsRequest conversion not implemented yet")
+            }
+
+            (ProviderRequestType::BedrockConverseStream(_), _) => {
+                todo!("Amazon Bedrock Stream to ChatCompletionsRequest conversion not implemented yet")
+            }
+
```
adilhafeez: why not?

crates/hermesllm/src/providers/response.rs
```diff
@@ -30,6 +34,8 @@ pub enum ProviderResponseType {
 pub enum ProviderStreamResponseType {
     ChatCompletionsStreamResponse(ChatCompletionsStreamResponse),
     MessagesStreamEvent(MessagesStreamEvent),
+    ConverseStreamEvent(ConverseStreamEvent),
+
```
adilhafeez:
```suggestion
```

crates/hermesllm/src/providers/response.rs
```diff
@@ -70,38 +75,172 @@ pub trait ProviderStreamResponse: Send + Sync {
...
+            ProviderStreamResponseType::ConverseStreamEvent(event) => {
+                // Use the Into<String> implementation for proper SSE formatting with event lines
+                event.into()
+            }
+            ProviderStreamResponseType::ChatCompletionsStreamResponse(_) => {
+                // For OpenAI, use simple data line format
+                let json = serde_json::to_string(&self).unwrap_or_default();
+                format!("data: {}\n\n", json)
+            }
+        }
```
adilhafeez, line 129: why can't we do event.into for chat completion streaming response?

crates/hermesllm/src/providers/response.rs
```diff
@@ -70,38 +75,172 @@ pub trait ProviderStreamResponse: Send + Sync {
...
+impl TryFrom<(&[u8], &SupportedAPIs, &ProviderId)> for ProviderResponseType {
+    type Error = std::io::Error;
+
+    fn try_from((bytes, client_api, provider_id): (&[u8], &SupportedAPIs, &ProviderId)) -> Result<Self, Self::Error> {
+        let upstream_api = provider_id.compatible_api_for_client(client_api, false);
+        match (&upstream_api, client_api) {
+            (SupportedUpstreamAPIs::OpenAIChatCompletions(_), SupportedAPIs::OpenAIChatCompletions(_)) => {
+                let resp: ChatCompletionsResponse = ChatCompletionsResponse::try_from(bytes)
+                    .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e))?;
+                Ok(ProviderResponseType::ChatCompletionsResponse(resp))
```
adilhafeez, line 148: this map_err boilerplate code could be better if you use ? for example,
```
let resp: ChatCompletionsResponse = ChatCompletionsResponse::try_from(bytes)?
```

but you'd need to return error that is acceptable

crates/hermesllm/src/providers/response.rs
```diff
@@ -207,177 +346,191 @@ impl Into<Vec<u8>> for SseEvent {
...
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "SSE parse error: {}", self.message)
+
+/// AWS Event Stream frame decoder wrapper
+pub struct BedrockBinaryFrameDecoder<B>
+where
+    B: Buf,
+{
+    decoder: aws_smithy_eventstream::frame::MessageFrameDecoder,
+    buffer: B,
```
adilhafeez:
```suggestion
    decoder: MessageFrameDecoder,
    buffer: B,
```

and import at file top aws_smithy_eventstream::frame::MessageFrameDecoder

crates/hermesllm/src/providers/response.rs
```diff
@@ -207,177 +346,191 @@ impl Into<Vec<u8>> for SseEvent {
...
+    }
 }
 
-impl fmt::Display for SseParseError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "SSE parse error: {}", self.message)
+
+/// AWS Event Stream frame decoder wrapper
+pub struct BedrockBinaryFrameDecoder<B>
+where
```
adilhafeez: so its not used in this file at all? I'd move it into a separate file for example bedrock_decoder.rs to keep this file simple

crates/hermesllm/src/transforms/request/from_anthropic.rs
adilhafeez, line 1: we shouldn't do refactoring and code changes in the same PR - it becomes difficult to track what code change was new vs a refactor

crates/hermesllm/src/lib.rs
```diff
@@ -74,4 +78,133 @@ mod tests {
...
+    /// 2. Handle chunked network arrivals with buffering
+    /// 3. Extract event types from message headers
+    /// 4. Parse JSON payloads from decoded messages
+    /// 5. Reconstruct streaming content from contentBlockDelta events
+    ///
+    /// The decoder handles frame boundaries automatically - you just keep calling
+    /// decode_frame() until it returns Incomplete, which means you've processed
+    /// all complete frames in the buffer.
+    #[test]
+    fn test_amazon_bedrock_streaming_response() {
```
adilhafeez, line 100: super amazing test !!!

crates/llm_gateway/src/stream_context.rs
```diff
@@ -45,8 +50,8 @@ pub struct StreamContext {
     traces_queue: Arc<Mutex<VecDeque<TraceData>>>,
     overrides: Rc<Option<Overrides>>,
     user_message: Option<String>,
-    /// Store upstream response status code to handle error responses gracefully
     upstream_status_code: Option<StatusCode>,
+    binary_frame_decoder: Option<BedrockBinaryFrameDecoder<bytes::BytesMut>>,
```
adilhafeez, line 56:
```suggestion
    bedrock_binary_frame_decoder: Option<BedrockBinaryFrameDecoder<bytes::BytesMut>>,
```

crates/llm_gateway/src/stream_context.rs
```diff
@@ -148,14 +155,19 @@ impl StreamContext {
...
                 self.set_http_request_header("anthropic-version", Some("2023-06-01"));
             }
-            Some(SupportedAPIs::OpenAIChatCompletions(_)) | None => {
+            Some(
+                SupportedUpstreamAPIs::OpenAIChatCompletions(_)
+                | SupportedUpstreamAPIs::AmazonBedrockConverse(_)
+                | SupportedUpstreamAPIs::AmazonBedrockConverseStream(_),
+            )
+            | None => {
                 // OpenAI and default: use Authorization Bearer token
```
adilhafeez, line 173: this whole thing can be replaced with `_`

crates/llm_gateway/src/stream_context.rs
```diff
@@ -487,6 +508,157 @@ impl StreamContext {
...
+        use hermesllm::providers::response::ProviderStreamResponseType;
+
+        // Initialize decoder if not present
+        if self.binary_frame_decoder.is_none() {
+            self.binary_frame_decoder = Some(BedrockBinaryFrameDecoder::from_bytes(&[]));
+        }
+
+        // Add incoming bytes to buffer
+        if let Some(decoder) = self.binary_frame_decoder.as_mut() {
+            decoder.buffer_mut().extend_from_slice(body);
```
adilhafeez: You could just get the decoder at the begining and then work on it e.g,

```
// note that unwrap will cause exception if decoder is none
let Some(decoder) = self.binary_frame_decoder.as_mut().unwrap();
```

doing so will make rest of the code a bit simpler where you don't have to do self.binary_frame_decoder.as_mut().unwrap() ...

crates/llm_gateway/src/stream_context.rs
```diff
@@ -487,6 +508,157 @@ impl StreamContext {
...
+                        Ok(provider_response) => {
+                            self.record_ttft_if_needed();
+
+                            // Extract index from the event if available
+                            let event_index =
+                                if let ProviderStreamResponseType::MessagesStreamEvent(ref evt) =
+                                    provider_response
+                                {
+                                    use hermesllm::apis::anthropic::MessagesStreamEvent;
+                                    match evt {
```
adilhafeez:
```suggestion
                                    use MessagesStreamEvent;
                                    match evt {
```

and import hermesllm::apis::anthropic::MessagesStreamEvent at the top of file

crates/llm_gateway/src/stream_context.rs
```diff
@@ -487,6 +508,157 @@ impl StreamContext {
...
+                                    }
+                                } else {
+                                    None
+                                };
+
+                            // Check event type to track ContentBlockStart
+                            if let Some(event_type) = provider_response.event_type() {
+                                match event_type {
+                                    "content_block_start" => {
+                                        // Mark that we've seen ContentBlockStart for this index
```
adilhafeez: uplift these static strings in consts

crates/hermesllm/src/providers/response.rs
```diff
@@ -207,177 +346,191 @@ impl Into<Vec<u8>> for SseEvent {
...
+    B: Buf,
+{
+    decoder: aws_smithy_eventstream::frame::MessageFrameDecoder,
+    buffer: B,
+    content_block_start_indices: std::collections::HashSet<i32>,
```
adilhafeez: ah this is for tracking content block start to fix their bug :(

crates/llm_gateway/src/stream_context.rs
```diff
@@ -487,6 +508,157 @@ impl StreamContext {
...
+                                        MessagesStreamEvent::ContentBlockStop { index, .. } => {
+                                            Some(*index as i32)
+                                        }
+                                        _ => None,
+                                    }
+                                } else {
+                                    None
+                                };
+
+                            // Check event type to track ContentBlockStart
```
adilhafeez: all this tracking for bedrock bug

arch/tools/cli/config_generator.py
salmanap: is that in the base_url? It would make our config more tricky to write and manage, and my sense is that the overall gain will be little.

crates/hermesllm/src/providers/response.rs
salmanap: yes - unfortunately.

crates/hermesllm/src/providers/request.rs
salmanap: Fair feedback - i'll open a new PR for renaming

crates/hermesllm/src/providers/request.rs
salmanap: because we don't support Amazon SDK as a client today

crates/hermesllm/src/providers/response.rs
salmanap: okay. will do

crates/hermesllm/src/providers/response.rs
salmanap: we should. for sure.

crates/hermesllm/src/providers/response.rs
salmanap: yea sure. this makes sense to me.

crates/hermesllm/src/lib.rs
salmanap: you got it.

crates/llm_gateway/src/stream_context.rs
salmanap: sure.

crates/llm_gateway/src/stream_context.rs
salmanap: I suppose that is fine.

crates/llm_gateway/src/stream_context.rs
salmanap: fair enough.

crates/common/src/consts.rs
salmanap: fixed.

arch/tools/cli/config_generator.py
salmanap: keeping - as that is consistent with together_ai azure_openai, etc

crates/hermesllm/src/clients/endpoints.rs
salmanap: fixed.

crates/hermesllm/src/clients/endpoints.rs
salmanap: fixed

crates/llm_gateway/src/stream_context.rs
salmanap: fixed. using Rust structs instead

crates/llm_gateway/src/stream_context.rs
salmanap: fixed.

crates/llm_gateway/src/stream_context.rs
salmanap: on closer inspection the challenge is that once we borrow a file in a struct, we can't refer to the struct until we are done borrowing. Meaning we can't call self.record_ttft_if_needed(); if we have a mutable field that we borrowed in the same scope. The fix will be a bit more involved and right now might not be worth the trouble.

arch/supervisord.conf
salmanap: fixed.

PR #591, opened 2025-10-14
add rag agent demo

adilhafeez:
sample screenshot

image


adilhafeez:
better demo https://github.com/katanemo/archgw/pull/621

PR #602, opened 2025-10-24
move pytest to dev deps and migrate to poetry 2

branchv:
First commit simply fixes some `pytest` dependencies from being installed at runtime.

Second commit migrates to [poetry 2](https://python-poetry.org/blog/announcing-poetry-2.0.0), which supports standard pep 517 `[project]` metadata. This helps #597 since now the only poetry-specific metadata left is the `include` for file selection, meaning you can now choose to use either `poetry` or `uv`.


salmanap:
This is great. Running workflows.

salmanap:
tests passed. https://github.com/katanemo/archgw/pull/603

PR #605, opened 2025-10-25
support python 3.14

branchv:
Not sure if `torch` needed to be pinned but 2.9 has 3.14 wheels. Also removed `setuptools` since I don't see it being used


branchvincent:
thank you too @salmanap!

PR #608, opened 2025-10-28
support base_url path for model providers

salmanap:


arch/tools/test/test_config_generator.py
```diff
@@ -243,14 +243,13 @@ def test_validate_and_render_happy_path_agent_config(monkeypatch):
...
-
   - model: custom/gpt-4o
 
 """,
     },
     {
-        "id": "base_url_no_prefix",
-        "expected_error": "Please provide base_url without path",
+        "id": "base_url_with_path_prefix",
+        "expected_error": None,
```
adilhafeez, line 252: should extend the test to ensure that `base_url_path_prefix` was set correctly with /api/v2

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -77,70 +77,91 @@ impl SupportedAPIs {
...
+        // Helper function to build endpoint with optional prefix override
+        let build_endpoint = |provider_prefix: &str, suffix: &str| -> String {
+            if !base_url_path_prefix.is_empty() {
+                // Use base_url_path_prefix instead of provider's default prefix
+                let prefix = base_url_path_prefix.trim_matches('/');
+                if prefix.is_empty() {
+                    // If prefix is just slashes, return suffix as-is
+                    suffix.to_string()
+                } else {
+                    format!("/{}{}", prefix, suffix)
```
adilhafeez: use Path::new(..).join(..)

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -77,70 +77,91 @@ impl SupportedAPIs {
...
+                }
+            } else {
+                // Use provider's default prefix
+                if provider_prefix.is_empty() {
+                    suffix.to_string()
+                } else {
+                    format!("{}{}", provider_prefix, suffix)
```
adilhafeez: same as above use `use Path::new(..).join(..)`

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -245,4 +266,327 @@ mod tests {
...
+                ""
+            ),
+            "/compatible-mode/v1/chat/completions"
+        );
+
+        // Test Azure OpenAI provider
+        assert_eq!(
+            api.target_endpoint_for_provider(
+                &ProviderId::AzureOpenAI,
+                "/v1/chat/completions",
```
adilhafeez, line 320: What if someone typed `/v1/chat/completions/`?

crates/hermesllm/src/clients/endpoints.rs
```diff
@@ -77,70 +77,91 @@ impl SupportedAPIs {
         request_path: &str,
         model_id: &str,
         is_streaming: bool,
+        base_url_path_prefix: &str,
```
adilhafeez: this should be called Optional so you don't have to compare with empty string

docs/source/concepts/llm_providers/supported_providers.rst
```diff
@@ -616,6 +618,92 @@ For providers that implement the OpenAI API but aren't natively supported:
...
+   * - ``https://api.z.ai/api/coding/paas/v4``
+     - ``/api/coding/paas/v4/chat/completions``
+   * - ``https://mycompany.openai.azure.com``
+     - ``/openai/deployments/{model}/...``
+   * - ``https://mycompany.openai.azure.com/custom/path``
+     - ``/custom/path/{model}/...``
+   * - ``http://localhost:11434`` (no path)
+     - ``/v1/chat/completions``
+   * - ``http://localhost:11434/api/v2``
+     - ``/api/v2/chat/completions``
```
adilhafeez: unnecessary pls remove

docs/source/concepts/llm_providers/supported_providers.rst
```diff
@@ -616,6 +618,92 @@ For providers that implement the OpenAI API but aren't natively supported:
...
+      - model: zhipu/glm-4.6
+        access_key: $ZHIPU_API_KEY
+        base_url: https://api.z.ai/api/coding/paas/v4
+        # Results in: https://api.z.ai/api/coding/paas/v4/chat/completions
+
+      # Azure with custom path
+      - model: azure_openai/gpt-4
+        access_key: $AZURE_API_KEY
+        base_url: https://mycompany.openai.azure.com/custom/deployment/path
+        # Path prefix /custom/deployment/path replaces /openai/deployments
```
adilhafeez: should continue to follow same example as above for example show `Results: ...`

adilhafeez (changes requested):

docs/source/concepts/llm_providers/supported_providers.rst
salmanap: if you look at the compiled docs, I think they are actual helpful for developers to see how the base_url would translate.

crates/hermesllm/src/clients/endpoints.rs
salmanap: For URL paths, we should either:

Keep using format! (which is actually fine and idiomatic for URL construction)
Use a URL manipulation crate like url
Since these are simple URL path concatenations and we don't need full URL parsing, format! is actually the most appropriate and idiomatic solution here. The std::path::Path crate is specifically for filesystem paths and would introduce incorrect behavior (e.g., on Windows it might use \ instead of /).

The feedback about using Path::new().join() would be correct for filesystem paths, but not for URL paths.

crates/hermesllm/src/clients/endpoints.rs
salmanap: The std::path::Path crate is specifically for filesystem paths and would introduce incorrect behavior (e.g., on Windows it might use \ instead of /).

crates/hermesllm/src/clients/endpoints.rs
salmanap: It would cause a 400. Then its no longer a base_url its a fully qualified path. We aren't deviating from the a norm here. Most egress functionality in a proxy/gateway supports this concepts especially given the usage pattern established by openai client libraries.

docs/source/concepts/llm_providers/supported_providers.rst
salmanap: fair. fixing.

crates/hermesllm/src/clients/endpoints.rs
salmanap: fixing.

docs/source/concepts/llm_providers/supported_providers.rst
salmanap: removed

arch/tools/test/test_config_generator.py
salmanap: this will require a big refactor. we do have a test for base_url not being set, especially in the instance where the model provider is custom.

PR #615, opened 2025-11-12
removing model_server python module to brightstaff (function calling)

salmanap:


arch/tools/cli/config_generator.py
```diff
@@ -304,6 +304,14 @@ def validate_and_render_schema():
...
 
+    # Always add arch-function model provider
+    updated_model_providers.append(
+        {
+            "name": "arch-function",
+            "provider_interface": "arch",
+            "model": "Arch-Function",
+        }
+    )
     config_yaml["model_providers"] = deepcopy(updated_model_providers)
```
adilhafeez: this is good, maybe we can also always add "arch-router" as well

arch/envoy.template.yaml
```diff
@@ -262,19 +262,16 @@ static_resources:
...
                             headers:
                               - name: "x-arch-upstream"
                                 string_match:
-                                  exact: {{ internal_cluster }}
+                                  exact: bright_staff
                           route:
                             auto_host_rewrite: true
-                            cluster: {{ internal_cluster }}
+                            cluster: bright_staff
                             timeout: 300s
```
adilhafeez, line 274: this whole block may not be needed

crates/hermesllm/src/apis/openai.rs
```diff
@@ -385,6 +385,8 @@ pub struct ChatCompletionsResponse {
     pub usage: Usage,
     pub system_fingerprint: Option<String>,
     pub service_tier: Option<String>,
+    // This isn't a standard OpenAI field, but we include it for extensibility
+    pub metadata: Option<HashMap<String, Value>>,
 }
```
adilhafeez, line 396: why add it here now?

crates/brightstaff/src/main.rs
```diff
@@ -125,7 +126,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
                     (&Method::POST, CHAT_COMPLETIONS_PATH | MESSAGES_PATH) => {
                         let fully_qualified_url =
                             format!("{}{}", llm_provider_url, req.uri().path());
-                        chat(req, router_service, fully_qualified_url, model_aliases)
+                        router_chat(req, router_service, fully_qualified_url, model_aliases)
                             .with_context(parent_cx)
                             .await
                     }
```
adilhafeez, line 132: while we're at it let's make default endpoint to router_chat like

```
(&Method::POST, _) => router_chat
````

File: crates/brightstaff/src/handlers/router.rs [hunk 5]
```diff
@@ -237,34 +234,12 @@ pub async fn chat(
         headers.insert(header_name, header_value.clone());
     }
 
-    // channel to create async stream
-    let (tx, rx) = mpsc::channel::<Bytes>(16);
+    // Use the streaming utility with a passthrough processor (no modification of chunks)
+    let byte_stream = llm_response.bytes_stream();
+    let processor = PassthroughProcessor;
+    let streaming_response = create_streaming_response(byte_stream, processor, 16);
```
adilhafeez, line 241: Nice abstraction but let's not create PassthroughProcessor right now until we need to modify the response. You can just name it sth like `create_pass_thorugh_response_processor`

crates/brightstaff/src/handlers/function_calling.rs
```diff
@@ -0,0 +1,1888 @@
...
+                let chunk = chunk.map_err(|e| FunctionCallingError::InvalidModelResponse(format!("Stream error: {}", e)))?;
+                if let Some(choice) = chunk.choices.first() {
+                    if let Some(content) = &choice.delta.content {
+                        model_response.push_str(content);
+                    }
+                }
+            }
+            info!("[Agent Orchestrator]: response received");
+        } else {
+            if let Some(tools) = request.tools.as_ref() {
```
adilhafeez, line 935: should remove agent orchestrator from here and from dmeos its adding complexity

crates/brightstaff/src/handlers/function_calling.rs
```diff
@@ -0,0 +1,1888 @@
...
+            llm_provider_url.clone(),
+        );
+        handler.function_handler.function_calling_chat(chat_request).await
+    } else {
+        let handler = ArchFunctionHandler::new(
+            chat_request.model.clone(),
+            ArchFunctionConfig::default(),
+            llm_provider_url.clone(),
+        );
+        handler.function_calling_chat(chat_request).await
```
adilhafeez, line 1319: we don't take in model name from user when making outbound call to arch-function

crates/brightstaff/src/handlers/function_calling.rs
adilhafeez, line 1: hallucination detection is a complicated code have you looked at it line by line and do you think we have enough tests to ensure that it works as expected?

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -371,7 +371,26 @@ impl StreamContext {
...
+                    })
+                    .collect();
+                Some(yaml_params)
+            },
+            Err(e) => {
+                warn!("Failed to parse tool call arguments: {}", e);
+                None
+            }
+        };
+
```
adilhafeez, line 394: I don't understand why is this needed now?

crates/prompt_gateway/src/stream_context.rs
```diff
@@ -870,7 +889,7 @@ mod test {
                         id: "1".to_string(),
                         function: common::api::open_ai::FunctionCallDetail {
                             name: "test".to_string(),
-                            arguments: None,
+                            arguments: "{}".to_string(),
```
adilhafeez, line 893: what if arguments were passed in a null or empty?

arch/tools/cli/config_generator.py
salmanap: Already there

        if (
            routing_model_provider is None
            and "arch-router" not in model_provider_name_set
        ):
            updated_model_providers.append(
                {
                    "name": "arch-router",
                    "provider_interface": "arch",
                    "model": config_yaml.get("routing", {}).get("model", "Arch-Router"),
                }
            )

arch/envoy.template.yaml
salmanap: for now we need to pass calls to arch_internal which then forwards calls to bright_staff. Removing the 11000 arch_internal cluster was a lot of refactoring. This was the simplest change. Happy to revisit and refactor futher.

crates/hermesllm/src/apis/openai.rs
salmanap: This is needed because Arch-Function returns a v1/chat/completions response and uses the metadata filed to store tool calls. In the python code, we had a custom version of the OpenAI chat completions object and modified it to support our use case. This is the same idea but now moving that logic to Rust to support our Arch-Function port to Rust.

crates/brightstaff/src/handlers/function_calling.rs
salmanap: Agree - but for this port I wanted to keep things as is. Tracking issue: https://github.com/katanemo/archgw/issues/617

crates/brightstaff/src/main.rs
salmanap: We already have a default (see below) And the challenge we have is that we only support two client endpoints: v1/chat/completions and v1/messages and if we get v1/responses which is the default for openAI clients they we will not handle that gracefully and barf. 

                    _ => {
                        debug!("No route for {} {}", req.method(), req.uri().path());
                        let mut not_found = Response::new(empty());
                        *not_found.status_mut() = StatusCode::NOT_FOUND;
                        Ok(not_found)
                    }

crates/brightstaff/src/handlers/router.rs
salmanap: That's a fair comment - and I really should fix in this PR. If you are okay with it, can we ship as is? To move this code out and test everybit of it was tricky to begin with and then i'll have to do the reverse. If it is not a blocker for you, can I address this in a separate PR at a later time?

crates/brightstaff/src/handlers/function_calling.rs
salmanap: Agreed - fixed. It will use a static constant.

crates/brightstaff/src/handlers/function_calling.rs
salmanap: I did look at it as deeply as I could.  And also chatted with @nehcgs and the feeling was the LLM-based line by line comparison we ddi via ClaudeCode was sufficient.

crates/prompt_gateway/src/stream_context.rs
salmanap: This conversion is necessary because the OpenAI API spec defines tools as a JSON string (for compatibility) and so when we serialize and deserialize using our new Rust constructs in heremes we lose the convenience of the python dictionary object that was being sent back by function_calling.py. 

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct FunctionCallDetail {
    pub name: String,
    pub arguments: String,
}

crates/prompt_gateway/src/stream_context.rs
salmanap: This is in a test where we are simulation presence of am "empty" argument dictionary as a string. Note that the upstrem LLM contract would break if it returns null or empty string. if there are no arguments to return its an empty dictionary. So we should be okay here.

crates/brightstaff/src/handlers/router.rs
adilhafeez: Not a blocker but let's take it up in the follow up PR

adilhafeez: let's create an issue at the very least to track this change

.github/workflows/e2e_tests.yml
```diff
@@ -14,6 +14,32 @@ jobs:
...
+          # sudo rm -rf /opt/hostedtoolcache || true
+
+          echo "=== Docker cleanup (before our builds/compose) ==="
+          docker system prune -af || true
+          docker volume prune -f || true
+
+          echo "=== Disk usage after cleanup ==="
+          df -h
+      # --- End disk cleanup section ---
+
```
adilhafeez, line 42: don't enable it by default

PR #621, opened 2025-11-24
Use mcp tools for filter chain

adilhafeez:


demos/use_cases/mcp_filter/arch_config.yaml
```diff
@@ -0,0 +1,41 @@
...
+    access_key: $OPENAI_API_KEY
+
+model_aliases:
+  fast-llm:
+    target: gpt-4o-mini
+  smart-llm:
+    target: gpt-4o
+
+listeners:
+  - type: agent
```
salmanap, line 30: is type and name absolutely required? I can see that you are probably future proofing listeners. But I would argue that is boilerplate that wouldn't add a lot of value to the first-run developer experience.  I am still of the belief that if you had it as prompt_targets, then you can have type:agent, which is super helpful then because then the container will call agents or tools/endpoints.

demos/use_cases/rag_agent/arch_config.yaml
```diff
@@ -0,0 +1,41 @@
+version: v0.3.0
+
+agents:
+  - id: rag_agent
+    url: http://host.docker.internal:10505
+
+agent_filters:
```
salmanap: I am leaning towards calling them just "filters". Its a bit more generic so could bite us both ways, but my sense is that just like agents are a unique entity, filters are unique in the way they are represented as a top-level object in our config.

demos/use_cases/mcp_filter/arch_config.yaml
```diff
@@ -0,0 +1,41 @@
+version: v0.3.0
+
+agents:
+  - id: rag_agent
+    url: http://host.docker.internal:10505
+
+agent_filters:
+  - id: query_rewriter
+    url: http://host.docker.internal:10501
+    # type: mcp # default is mcp
```
salmanap, line 10: I would lead with our MCP example in the example use case.

arch/tools/cli/config_generator.py
```diff
@@ -101,8 +101,17 @@ def validate_and_render_schema():
 
     # Process agents section and convert to endpoints
     agents = config_yaml.get("agents", [])
-    for agent in agents:
+    agent_filters = config_yaml.get("agent_filters", [])
```
salmanap: please see comment on "filters" below.

demos/use_cases/mcp_filter/src/rag_agent/__init__.py
salmanap, line 1: My sense is that we should rename this folder to demos/use_cases/mcp_filters - I think the core functionality we are demonstrating here is executing a filter chain via MCP and returning the response from an agent. We could have a separate demo for multi_agent_orchestration

crates/brightstaff/src/main.rs
```diff
@@ -60,8 +60,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        .as_deref()
+        .unwrap_or_default()
+        .iter()
+        .chain(arch_config.agent_filters.as_deref().unwrap_or_default())
+        .cloned()
+        .collect();
+
     let llm_providers = Arc::new(RwLock::new(arch_config.model_providers.clone()));
-    let agents_list = Arc::new(RwLock::new(arch_config.agents.clone()));
+    let agents_list = Arc::new(RwLock::new(Some(all_agents)));
```
salmanap: for easier code maintenance we should name this combined_agents_filters_list. Else someone will have to backtrack to the all_agents code understand that we are merging the list. The new variable name will stop them in their tracks and validate their assumptions better

crates/hermesllm/src/providers/request.rs
```diff
@@ -49,6 +49,362 @@ pub trait ProviderRequest: Send + Sync {
     fn get_temperature(&self) -> Option<f32>;
 }
 
+impl ProviderRequestType {
+    /// Get message history as OpenAI Message format
+    /// This is useful for processing chat history across different provider formats
+    pub fn get_message_history(&self) -> Vec<crate::apis::openai::Message> {
```
salmanap: This code path is a bit hard to follow - I would prefer this function to be in the trait and have its implementation guided via TryFrom blocks. This way they are contained in the right structs, and much easier to update and unit test. 

Plus there is already transformation logic in to_openai_responses.rs. This way we aren't writing new pieces of code on transformation. Else we will likely get it wrong and have downstream effects.

For e.g.


```
    fn extract_messages_text(&self) -> String {
        match self {
            Self::ChatCompletionsRequest(r) => r.extract_messages_text(),
            Self::MessagesRequest(r) => r.extract_messages_text(),
            Self::BedrockConverse(r) => r.extract_messages_text(),
            Self::BedrockConverseStream(r) => r.extract_messages_text(),
            Self::ResponsesAPIRequest(r) => r.extract_messages_text(),
        }
    }
```

crates/brightstaff/src/handlers/agent_chat_completions.rs
salmanap, line 1: shouldn't we just call this agent_invoke.rs and have it support multiple clients like /v1/responses /v1/messages and v1/chat/completions?

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -140,42 +160,64 @@ async fn handle_agent_chat(
...
+    //             "Failed to parse request body as ChatCompletionsRequest: {}",
+    //             err
+    //         );
+    //         AgentFilterChainError::RequestParsing(err)
+    //     })?;
+
     // Extract trace parent for routing
     let trace_parent = request_headers
         .iter()
         .find(|(key, _)| key.as_str() == "traceparent")
```
salmanap: please don't use string. Instead this is in consts.rs. Nit.

crates/brightstaff/src/handlers/agent_selector.rs
```diff
@@ -131,20 +137,23 @@ impl AgentSelector {
...
-                model: agent.id.clone(),
+        let mut preferences = Vec::new();
+
+        for agent_chain in agents {
+            preferences.push(ModelUsagePreference {
+                model: agent_chain.id.clone(),
                 routing_preferences: vec![RoutingPreference {
-                    name: agent.id.clone(),
-                    description: agent.description.as_ref().unwrap_or(&String::new()).clone(),
+                    name: agent_chain.id.clone(),
```
salmanap, line 150: These descriptions don't come from MCP today. Correct? Is that through a different PR?

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -53,18 +70,110 @@ impl PipelineProcessor {
...
+        agent_name: &str,
+        tool_name: &str,
+        start_time: SystemTime,
+        end_time: SystemTime,
+        elapsed: std::time::Duration,
+    ) {
+        let (trace_id, parent_span_id) = self.extract_trace_context();
+
+        let mut span_builder = SpanBuilder::new(format!("filter_execution: {}", agent_name))
+            .with_kind(SpanKind::Internal)
```
salmanap: This should be SpanKind::Client. Because we are making an outbound call to an MCP server.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -53,18 +70,110 @@ impl PipelineProcessor {
...
+            .with_start_time(start_time)
+            .with_end_time(end_time)
+            .with_attribute("filter_name", agent_name.to_string())
```
salmanap: call this "filter" - because that is how its represented in the arch_confing.yaml. The closer we are to the developer experience. The better it is.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -53,18 +70,110 @@ impl PipelineProcessor {
...
+        collector: &std::sync::Arc<common::traces::TraceCollector>,
+        agent_name: &str,
+        tool_name: &str,
+        start_time: SystemTime,
+        end_time: SystemTime,
+        elapsed: std::time::Duration,
+    ) {
+        let (trace_id, parent_span_id) = self.extract_trace_context();
+
+        let mut span_builder = SpanBuilder::new(format!("filter_execution: {}", agent_name))
```
salmanap: Please use the OperationNameBuilder utility to construct the operation names of your Spans. Operations should be {METHOD}{PATH}{MODEL/AGENT/FILTER}. The service name will be plano(agent) and underneath that it will have multiple spans for plano(filter) as the service name and each 

Please see tracing/constants.rs


`
/// Canonical operation name components for Arch Gateway
pub mod operation_component {
    /// Inbound request handling
    pub const INBOUND: &str = "plano(inbound)";

    /// Routing decision phase
    pub const ROUTING: &str = "plano(routing)";

    /// Handoff to upstream service
    pub const HANDOFF: &str = "plano(handoff)";

    /// Agent filter execution
    pub const AGENT_FILTER: &str = "plano(agent filter)";

    /// Agent execution
    pub const AGENT: &str = "plano(agent)";

    /// LLM call
    pub const LLM: &str = "plano(llm)";
}
`

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -53,18 +70,110 @@ impl PipelineProcessor {
...
+
+        if !trace_id.is_empty() {
+            span_builder = span_builder.with_trace_id(trace_id);
+        }
+        if let Some(parent_id) = parent_span_id {
+            span_builder = span_builder.with_parent_span_id(parent_id);
+        }
+
+        let span = span_builder.build();
+        collector.record_span("brightstaff", span);
```
salmanap: please use the service name via constants. This way we know what we can update it once it will apply neatly. For all agent calls it should be imho plano(agent) -> plano(fitler)... plano(invoke)

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -53,18 +70,110 @@ impl PipelineProcessor {
...
+    }
+
+    /// Record a span for MCP protocol interactions
+    fn record_mcp_span(
```
salmanap, line 135: Don't follow this vs record_filter_span.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -73,119 +182,452 @@ impl PipelineProcessor {
...
+                "mcp-session-id",
+                hyper::header::HeaderValue::from_str(sid).unwrap(),
+            );
+        }
+
+        Ok(headers)
+    }
+
+    /// Parse SSE formatted response and extract JSON-RPC data
+    fn parse_sse_response(&self, response_bytes: &[u8], agent_id: &str) -> Result<String, PipelineError> {
```
salmanap: We have an SSeStreamIter and SSeEvent structs in hermes that encapsulate this functionality. Any reason we aren't using that to process SSeEvents?

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -73,119 +182,452 @@ impl PipelineProcessor {
...
+        let arguments = serde_json::json!({
+            "messages": messages
+        });
+
+        let params = serde_json::json!({
+            "name": tool_name,
+            "arguments": arguments
+        });
+
+        Ok(JsonRpcRequest {
```
salmanap, line 400: If we will always use 2.0. Then we should serialize in the JsonRpcRequest. Else other callers will have to remember this, when that is the default

crates/brightstaff/src/handlers/jsonrpc.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum JsonRpcId {
+  String(String),
+  Number(u64),
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JsonRpcRequest {
```
salmanap, line 17: JsonRPCRequest?

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -73,119 +182,452 @@ impl PipelineProcessor {
...
+        Ok(response)
+    }
+
+    /// Build a tools/call JSON-RPC request
+    fn build_tool_call_request(
+        &self,
+        tool_name: &str,
+        messages: &[Message],
+    ) -> Result<JsonRpcRequest, PipelineError> {
+        let arguments = serde_json::json!({
```
salmanap: Why don't we build a TryFrom so that its more idiomatic?

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -73,119 +182,452 @@ impl PipelineProcessor {
...
+        }
 
-            if status.is_client_error() {
-                // 4xx errors - cascade back to developer
-                return Err(PipelineError::ClientError {
+        // Handle HTTP errors
+        if !http_status.is_success() {
+            let error_body = String::from_utf8_lossy(&response_bytes).to_string();
+            return Err(if http_status.is_client_error() {
+                PipelineError::ClientError {
```
salmanap, line 493: If one of the filter fails, do we throw this up and terminate the request? A bit harder to navigate the code path in web code review mode.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -73,119 +182,452 @@ impl PipelineProcessor {
...
+        self.send_initialized_notification(agent_id, &session_id, trace_collector)
+            .await
+            .expect("Failed to send initialized notification");
+
+        session_id
     }
 
     /// Send request to terminal agent and return the raw response for streaming
-    pub async fn invoke_upstream_agent(
+    pub async fn invoke_terminal_agent(
```
salmanap: we should just call this invoke_agent. As there is no concept of "terminal". We are invoking the agent and its filters.

demos/use_cases/mcp_filter/src/rag_agent/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+    default="streamable-http",
+    help="Transport type: stdio or sse",
+)
+@click.option("--host", "host", default="localhost", help="Host to bind MCP server to")
+@click.option("--port", "port", type=int, default=10500, help="Port for MCP server")
+@click.option(
+    "--agent",
+    "agent",
+    required=True,
+    help="Agent name: query_rewriter, context_builder, or response_generator",
```
salmanap, line 20: Are we mixing agents and filters in the MCP server. That will be confusing for the reader, because we want to drive a distinction between an agent that does "work" and filters are intended to "mutate" requests. Plus in our config they are separate things, and in this example they are combinned.

demos/use_cases/mcp_filter/src/rag_agent/api.py
```diff
@@ -0,0 +1,36 @@
+from pydantic import BaseModel
```
salmanap, line 1: I think pydantic has shapes for ChatCompletions. That would be more idiomatic to show to python developers

salmanap (commented):
Have a few comments. Feel free to call me so that we can start resolving these. Major themes

1. Trace building isn't ideal. Don't understand the diff between MCP spans and others.
2. We are re-building alot of the sse parsing logic by hand again
3. We don't seem to be getting the agent description via MCP (i couldn't tell if we were). 
4. Config suggestions.

demos/use_cases/mcp_filter/src/rag_agent/__init__.py
adilhafeez: renamed

crates/hermesllm/src/providers/request.rs
adilhafeez: will do - that is a good feedback

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: updated - thanks!

crates/brightstaff/src/handlers/agent_selector.rs
adilhafeez: Not yet - in follow up PR

crates/brightstaff/src/handlers/jsonrpc.rs
adilhafeez: I could but in my personal preference JsonRpcRequest flows better :)

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: I tried - it seems like serialization would be tricky and would require multiple rounds of serialization/deserialization to work and it will be sub-optimal. I am going to keep it as it is for now. But I did update the tool call and version to be consts.

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: yes we will do early terminate and clients will be able to handle that

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: updated to use const

crates/hermesllm/src/providers/request.rs
adilhafeez: updated for both get and set messages

arch/tools/cli/config_generator.py
adilhafeez: updated

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: That is a decent suggestion but I don't want to take it in this PR

crates/hermesllm/src/providers/request.rs
adilhafeez: updated

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: updated

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: its fixed now

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: updated

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: updated

crates/brightstaff/src/handlers/pipeline_processor.rs
adilhafeez: this is not json chat_completions its using a different pattern and sse code that we have may not be useful here

demos/use_cases/mcp_filter/arch_config.yaml
adilhafeez: Yes, it is needed in other parts of the system too. I'll keep it here for now and we can revisit it later.

demos/use_cases/rag_agent/arch_config.yaml
adilhafeez: updated

demos/use_cases/mcp_filter/arch_config.yaml
adilhafeez: updated

demos/use_cases/mcp_filter/src/rag_agent/api.py
adilhafeez: its good change - but not in this PR - let's take it up in follow up PR

PR #622, opened 2025-11-24
Add support for v1/responses API

salmanap:
No persistent storage yet - there will be a follow up PR for that

### DIFF
--- a/crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
+++ b/crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
@@ -24,6 +24,9 @@ pub struct AnthropicMessagesStreamBuffer {
     /// Track if we need to inject ContentBlockStop before message_delta
     needs_content_block_stop: bool,
 
+    /// Track if we've seen a MessageDelta (so we need to send MessageStop at the end)
+    seen_message_delta: bool,
+
     /// Model name to use when generating message_start events
     model: Option<String>,
 }
@@ -35,6 +38,7 @@ impl AnthropicMessagesStreamBuffer {
             message_started: false,
             content_block_start_indices: HashSet::new(),
             needs_content_block_stop: false,
+            seen_message_delta: false,
             model: None,
         }
     }
@@ -182,7 +186,7 @@ impl SseStreamBufferTrait for AnthropicMessagesStreamBuffer {
                         // Content deltas are between ContentBlockStart and ContentBlockStop
                         self.buffered_events.push(event);
                     }
-                    MessagesStreamEvent::MessageDelta { .. } => {
+                    MessagesStreamEvent::MessageDelta { usage, .. } => {
                         // Inject ContentBlockStop before message_delta
                         if self.needs_content_block_stop {
                             let content_block_stop = AnthropicMessagesStreamBuffer::create_content_block_stop_event();
@@ -190,11 +194,44 @@ impl SseStreamBufferTrait for AnthropicMessagesStreamBuffer {
                             self.needs_content_block_stop = false;
                         }
 
-                        // Add the message_delta event
+                        // Check if the last event was also a MessageDelta - if so, merge them
+                        // This handles Bedrock's split of stop_reason (MessageStop) and usage (Metadata)
+                        if let Some(last_event) = self.buffered_events.last_mut() {
+                            if let Some(ProviderStreamResponseType::MessagesStreamEvent(
+                                MessagesStreamEvent::MessageDelta {
+                                    usage: last_usage,
+                                    ..
+                                }
+                            )) = &mut last_event.provider_stream_response {
+                                // Merge: take stop_reason from first, usage from second (if non-zero)
+                                if usage.input_tokens > 0 || usage.output_tokens > 0 {
+                                    *last_usage = usage.clone();
+                                }
+                                // Mark that we've seen MessageDelta (need to send MessageStop later)
+                                self.seen_message_delta = true;
+                                // Don't push the new event, we've merged it
+                                return;
+                            }
+                        }
+
+                        // No previous MessageDelta to merge with, add this one
+                        self.buffered_events.push(event);
+                        self.seen_message_delta = true;
+                    }
+                    MessagesStreamEvent::ContentBlockStop { .. } => {
+                        // ContentBlockStop received from upstream (e.g., Bedrock)
+                        // Clear the flag so we don't inject another one
+                        self.needs_content_block_stop = false;
+                        self.buffered_events.push(event);
+                    }
+                    MessagesStreamEvent::MessageStop => {
+                        // MessageStop received from upstream (e.g., OpenAI via [DONE])
+                        // Clear the flag so we don't inject another one
+                        self.seen_message_delta = false;
                         self.buffered_events.push(event);
                     }
                     _ => {
-                        // Other Anthropic event types (ContentBlockStop, MessageStop, etc.), just accumulate
+                        // Other Anthropic event types (Ping, etc.), just accumulate
                         self.buffered_events.push(event);
                     }
                 }
@@ -207,14 +244,26 @@ impl SseStreamBufferTrait for AnthropicMessagesStreamBuffer {
     }
 
     fn into_bytes(&mut self) -> Vec<u8> {
-        // Inject ContentBlockStop if needed before flushing
-        if self.needs_content_block_stop {
-            let content_block_stop = AnthropicMessagesStreamBuffer::create_content_block_stop_event();
-            self.buffered_events.push(content_block_stop);
-            self.needs_content_block_stop = false;
+        // Convert all accumulated events to bytes and clear buffer
+        // NOTE: We do NOT inject ContentBlockStop here because it's injected when we see MessageDelta
+        // or MessageStop. Injecting it here causes premature ContentBlockStop in the middle of streaming.
+
+        // Inject MessageStop after MessageDelta if we've seen one
+        // This completes the Anthropic Messages API event sequence
+        if self.seen_message_delta {
+            let message_stop = MessagesStreamEvent::MessageStop;
+            let sse_string: String = message_stop.into();
+            let message_stop_event = SseEvent {
+                data: None,
+                event: Some("message_stop".to_string()),
+                raw_line: sse_string.clone(),
+                sse_transformed_lines: sse_string,
+                provider_stream_response: None,
+            };
+            self.buffered_events.push(message_stop_event);
+            self.seen_message_delta = false;
         }
 
-        // Convert all accumulated events to bytes and clear buffer
         let mut buffer = Vec::new();
         for event in self.buffered_events.drain(..) {
             let event_bytes: Vec<u8> = event.into();
@@ -344,10 +393,10 @@ data: {"id":"chatcmpl-456","object":"chat.completion.chunk","created":1234567890
         assert!(output.contains("\"text\":\" in San Francisco\""), "Should have second content delta");
         assert!(output.contains("\"text\":\" is\""), "Should have third content delta");
 
-        // For partial streams, the buffer will inject content_block_stop in into_bytes()
-        // because needs_content_block_stop is true. This is expected behavior to maintain
-        // proper Anthropic protocol even for incomplete streams.
-        assert!(output.contains("event: content_block_stop"), "Should have content_block_stop (injected at flush)");
+        // For partial streams (no finish_reason, no [DONE]), we do NOT inject content_block_stop
+        // because the stream may continue. This is correct behavior - only inject lifecycle events
+        // when we have explicit signals from upstream (finish_reason, [DONE], etc.)
+        assert!(!output.contains("event: content_block_stop"), "Should NOT have content_block_stop for partial stream");
 
         // Should NOT have completion events
         assert!(!output.contains("event: message_delta"), "Should NOT have message_delta");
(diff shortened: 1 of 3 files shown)


adilhafeez:
pls update the description and title

crates/brightstaff/src/main.rs
```diff
@@ -123,7 +123,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
 
             async move {
                 match (req.method(), req.uri().path()) {
-                    (&Method::POST, CHAT_COMPLETIONS_PATH | MESSAGES_PATH) => {
+                    (&Method::POST, CHAT_COMPLETIONS_PATH | MESSAGES_PATH | OPENAI_RESPONSES_API_PATH) => {
```
adilhafeez, line 126: Why prefix with openai when other path don't have openai in them?

consider changing OPENAI_RESPONSES_API_PATH => RESPONSES_API_PATH

salmanap: responses is specific to OpenAI while chat/completions is supported by many upstream providers like Grok. The one thing we can do, however, is to rename MESSAGES_PATH to ANTHROPIC_MESSAGES_API_PATH to be more consistent.

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
```diff
@@ -0,0 +1,447 @@
...
+
+        SseEvent {
+            data: None,
+            event: Some("content_block_stop".to_string()),
+            raw_line: sse_string.clone(),
+            sse_transformed_lines: sse_string,
+            provider_stream_response: None,
+        }
+    }
+}
```
adilhafeez, line 119: These methods doesn't need to part of &self as they are don't depend on any of internal state. Static methods make testing simpler and code more readble.

and then you could use sth like this to call create_content_block_start_event `AnthropicMessagesStreamBuffer::create_content_block_start_event`

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
```diff
@@ -0,0 +1,447 @@
...
+        // Skip ping messages
+        if event.should_skip() {
+            return;
+        }
+
+        // FIRST: Try to extract model name from the raw event data before transformation
+        // The provider_stream_response has already been transformed to Anthropic format,
+        // so we need to extract the model from the original raw data if available
+        if self.model == "unknown" {
+            if let Some(data) = &event.data {
```
adilhafeez: "unknown" is not a great way to test that "model" was not set. Use optional instead sth like this `model: Option<String>`.

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
```diff
@@ -0,0 +1,447 @@
...
+                    _ => {
+                        // Other event types, just accumulate
+                        self.buffered_events.push(event);
+                        return;
+                    }
+                }
+            }
+        }
+
+        // For events without provider_stream_response or event_type, just accumulate
```
adilhafeez:
```suggestion
        // For events without provider_stream_response or event_type, just accumulate the event
```

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
```diff
@@ -0,0 +1,447 @@
...
+                        return;
+                    }
+                    _ => {
+                        // Other event types, just accumulate
+                        self.buffered_events.push(event);
+                        return;
+                    }
+                }
+            }
+        }
```
adilhafeez: instead of adding `return` statement in every block just return at the end of this if block `if let Some(event_type) = provider_response.event_type() {`

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
```diff
@@ -0,0 +1,447 @@
...
+        for event in self.buffered_events.drain(..) {
+            let event_bytes: Vec<u8> = event.into();
+            buffer.extend_from_slice(&event_bytes);
+        }
+        buffer
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
adilhafeez, line 277: can you add tests to cover case where `provider_response` is not set so we ensure that passthrough case is working too?

crates/hermesllm/src/apis/streaming_shapes/chat_completions_streaming_buffer.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn into_bytes(&mut self) -> Vec<u8> {
+        // No finalization needed for OpenAI Chat Completions
+        // The [DONE] marker is already handled by the transformation layer
+        let mut buffer = Vec::new();
+        for event in self.buffered_events.drain(..) {
+            let event_bytes: Vec<u8> = event.into();
+            buffer.extend_from_slice(&event_bytes);
+        }
+        buffer
+    }
```
adilhafeez, line 38: just to make sure we're clear that this `drain` method will mutate and remove all elements from `buffered_events` and that this method cannot be called again. I think for cases like this you should throw an error if someone accidentally called `into_bytes` after it has already been consumed.

crates/hermesllm/src/apis/streaming_shapes/passthrough_streaming_buffer.rs
```diff
@@ -0,0 +1,95 @@
...
+        // Skip ping messages
+        if event.should_skip() {
+            return;
+        }
+
+        // Skip events with empty transformed lines (e.g., suppressed event-only lines)
+        if event.sse_transformed_lines.is_empty() {
+            return;
+        }
+
```
adilhafeez, line 28: For passthrough should we just send everything? Why are we filtering out ping and empty lines?

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
```diff
@@ -0,0 +1,590 @@
...
+        ResponsesAPIStreamEvent::ResponseCreated { .. } => "response.created",
+        ResponsesAPIStreamEvent::ResponseInProgress { .. } => "response.in_progress",
+        ResponsesAPIStreamEvent::ResponseCompleted { .. } => "response.completed",
+        ResponsesAPIStreamEvent::ResponseOutputItemAdded { .. } => "response.output_item.added",
+        ResponsesAPIStreamEvent::ResponseOutputItemDone { .. } => "response.output_item.done",
+        ResponsesAPIStreamEvent::ResponseOutputTextDelta { .. } => "response.output_text.delta",
+        ResponsesAPIStreamEvent::ResponseOutputTextDone { .. } => "response.output_text.done",
+        ResponsesAPIStreamEvent::ResponseFunctionCallArgumentsDelta { .. } => "response.function_call_arguments.delta",
+        ResponsesAPIStreamEvent::ResponseFunctionCallArgumentsDone { .. } => "response.function_call_arguments.done",
+        _ => "unknown",
```
adilhafeez: we should emit error log or fail if unknown even is encountered

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
```diff
@@ -0,0 +1,590 @@
...
+    };
+
+    let json_data = serde_json::to_string(&event).unwrap_or_default();
+    let wire_format: String = event.into();
```
adilhafeez: so this will fail silently - we should log error in case of any failure and not eat them silently

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
```diff
@@ -0,0 +1,590 @@
...
+
+    SseEvent {
+        data: Some(json_data),
+        event: Some(event_type.to_string()),
+        raw_line: wire_format.clone(),
+        sse_transformed_lines: wire_format,
+        provider_stream_response: None,
```
adilhafeez, line 41: why send duplicate data? it would take up double the state

crates/hermesllm/src/apis/streaming_shapes/sse.rs
```diff
@@ -1,10 +1,73 @@
...
+    fn into_bytes(&mut self) -> Vec<u8> {
+        match self {
+            Self::Passthrough(buffer) => buffer.into_bytes(),
+            Self::OpenAIChatCompletions(buffer) => buffer.into_bytes(),
+            Self::AnthropicMessages(buffer) => buffer.into_bytes(),
+            Self::OpenAIResponses(buffer) => buffer.into_bytes(),
+        }
+    }
+}
+
```
adilhafeez, line 70: all of this could be rewritten as
```
impl SseStreamBufferTrait for SseStreamBuffer {
    fn add_transformed_event(&mut self, event: SseEvent) {
      buffer.add_transformed_event(event)
    }

    fn into_bytes(&mut self) -> Vec<u8> {
        self.into_bytes()
    }
}
```

crates/llm_gateway/src/stream_context.rs
```diff
@@ -527,12 +542,15 @@ impl StreamContext {
...
+                    self.sse_buffer
+                        .as_mut()
+                        .unwrap()
+                        .add_transformed_event(transformed_event);
                 }
 
-                Ok(response_buffer)
+                // Get accumulated bytes from buffer and return
+                Ok(self.sse_buffer.as_mut().unwrap().into_bytes())
             }
```
adilhafeez, line 571: `.unwrap()` could cause panic. I think you can work off from sse_buffer var you create from line 484 and then use that var at like 545 and 552.

adilhafeez (commented):
added some feedback - pls take a look at it and let me know what are your thought on them

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
salmanap: Agree. Fixed

salmanap: Although in the responses_api_streaming_buffer.rs there is state attached to &self like (response_id, model, sequence_number)

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
salmanap: fixed

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
salmanap: fixed

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
salmanap: If we assign wire_format to both the variables, we will have a borrow issue. In order to avoid borrow issues we have to clone. The other way to do this to optimize memory usage is to use Arc<String> type so that we have pointers to the same object. Not sure if that optimization refactoring is worth it as the events are short lived.

crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
salmanap: that test exist via the PassthroughStreamBuffer. When we detect that the client api and upstream api are the same. We don't create an AnthropicStreamBuffer. That logic is in streamin_responses.rs

impl TryFrom<(&SupportedAPIsFromClient, &SupportedUpstreamAPIs)>
    for SseStreamBuffer
{
    type Error = Box<dyn std::error::Error + Send + Sync>;

    fn try_from(
        (client_api, upstream_api): (&SupportedAPIsFromClient, &SupportedUpstreamAPIs),
    ) -> Result<Self, Self::Error> {

        // If APIs match, use passthrough - no buffering/transformation needed
        if !needs_buffering(client_api, upstream_api) {
            return Ok(SseStreamBuffer::Passthrough(PassthroughStreamBuffer::new()));
        }

        // APIs differ - use appropriate buffer for client API
        match client_api {
            SupportedAPIsFromClient::OpenAIChatCompletions(_) => {
                Ok(SseStreamBuffer::OpenAIChatCompletions(OpenAIChatCompletionsStreamBuffer::new()))
            }
            SupportedAPIsFromClient::AnthropicMessagesAPI(_) => {
                Ok(SseStreamBuffer::AnthropicMessages(AnthropicMessagesStreamBuffer::new()))
            }
            SupportedAPIsFromClient::OpenAIResponsesAPI(_) => {
                Ok(SseStreamBuffer::OpenAIResponses(ResponsesAPIStreamBuffer::new()))
            }
        }
    }
}

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
salmanap: fixed

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
salmanap: logging error. if we can't find a non known that we can convert to, we default to 'unknown' for now, I think you are right we need to know what is the right failure scenario - should we throw that upstream or just log it. Meaning if the content delta of the event can be parsed by the client should we consider this a failure, even if we aren't able to generate the right event? Don't know.

crates/llm_gateway/src/stream_context.rs
salmanap: fixed.

adilhafeez (approved):

PR #627, opened 2025-12-09
handle agent error better

adilhafeez:
if agent filter return 4xx, properly cascade the error text back to the developer

this code change will help build agent filters and allow them to early terminate if for example intent deviates or jailbreak attempt is made

### DIFF
--- a/crates/brightstaff/src/handlers/agent_chat_completions.rs
+++ b/crates/brightstaff/src/handlers/agent_chat_completions.rs
@@ -59,8 +59,8 @@ pub async fn agent_chat(
 
                 let json_string = error_json.to_string();
                 let mut response = Response::new(ResponseHandler::create_full_body(json_string));
-                *response.status_mut() =
-                    hyper::StatusCode::from_u16(*status).unwrap_or(hyper::StatusCode::BAD_REQUEST);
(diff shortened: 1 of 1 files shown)


crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -37,7 +37,38 @@ pub async fn agent_chat(
...
+                    "agent": agent,
+                    "status": status,
+                    "agent_response": body
+                });
+
+                let json_string = error_json.to_string();
+                let mut response = Response::new(ResponseHandler::create_full_body(json_string));
+                *response.status_mut() =
+                    hyper::StatusCode::from_u16(*status).unwrap_or(hyper::StatusCode::BAD_REQUEST);
+                response.headers_mut().insert(
```
salmanap, line 64: shouldn't we send the status code sent from the agent/filter. Any reason we want to mask that into a 400?

adilhafeez: That is good observation - for client errors that we receive from filter we should just pass it through. Fixed it in new code.

salmanap (approved):

PR #628, opened 2025-12-10
Improve end to end tracing

salmanap:
Moving tracing out of the WASM filter and intro brightstaff to write even higher-fidelity traces and spans. By doing such, we have consolidate all request path via brightstaff.

### DIFF
--- a/crates/brightstaff/src/handlers/llm.rs
+++ b/crates/brightstaff/src/handlers/llm.rs
@@ -114,12 +114,6 @@ pub async fn llm_chat(
         full_qualified_llm_provider_url, model_name
     );
 
-    // Extract trace_parent for upstream request headers
-    let trace_parent = request_headers
-        .iter()
-        .find(|(ty, _)| ty.as_str() == "traceparent")
-        .map(|(_, value)| value.to_str().unwrap_or_default().to_string());
-
     request_headers.insert(
         ARCH_PROVIDER_HINT_HEADER,
         header::HeaderValue::from_str(&model_name).unwrap(),
@@ -129,13 +123,6 @@ pub async fn llm_chat(
         header::HeaderName::from_static(ARCH_IS_STREAMING_HEADER),
         header::HeaderValue::from_str(&is_streaming_request.to_string()).unwrap(),
     );
-
-    if let Some(trace_parent) = trace_parent {
-        request_headers.insert(
-            header::HeaderName::from_static("traceparent"),
-            header::HeaderValue::from_str(&trace_parent).unwrap(),
-        );
-    }
     // remove content-length header if it exists
     request_headers.remove(header::CONTENT_LENGTH);
 


arch/envoy.template.yaml
```diff
@@ -51,11 +51,11 @@ static_resources:
                         envoy_grpc:
                           cluster_name: opentelemetry_collector
                         timeout: 0.250s
-                      service_name: archgw(inbound)
+                      service_name: plano(inbound)
                   random_sampling:
                     value: {{ arch_tracing.random_sampling }}
                 {% endif %}
-                stat_prefix: ingress_traffic
+                stat_prefix: plano(inbound)
```
adilhafeez, line 58: do we want to name it sth that conveys that it is actually about prompt gateway like sth maybe plano(inbound, prompt)

arch/envoy.template.yaml
```diff
@@ -240,7 +225,7 @@ static_resources:
                         envoy_grpc:
                           cluster_name: opentelemetry_collector
                         timeout: 0.250s
-                      service_name: outbound_api_traffic
+                      service_name: tool
                   random_sampling:
```
adilhafeez, line 229: do we want to keep this listener internal, if so then maybe let's just bind it to 127.0.0.1

arch/envoy.template.yaml
```diff
@@ -413,7 +398,7 @@ static_resources:
                         envoy_grpc:
                           cluster_name: opentelemetry_collector
                         timeout: 0.250s
-                      service_name: archgw(outbound)
+                      service_name: plano(outbound)
```
adilhafeez, line 401: it will always be llm traffic should it be called `plano(llm)` as thats more clear or maybe `plano(outbound, llm)`

arch/envoy.template.yaml
```diff
@@ -484,6 +469,50 @@ static_resources:
...
+                              num_retries: 3
+                              per_try_timeout: 2s
+                              host_selection_retry_max_attempts: 5
+                              retriable_status_codes: [500, 502, 503, 504]
+                http_filters:
+                  - name: envoy.filters.http.router
+                    typed_config:
+                      "@type": type.googleapis.com/envoy.extensions.filters.http.router.v3.Router
+{% endif %}
+
```
adilhafeez, line 515: You don't really need this you could directly send traces to opentelemetry_collector_http, this entire listener is not necessary

crates/brightstaff/src/handlers/router.rs
```diff
@@ -11,25 +10,40 @@ use hyper::header::{self};
...
-    let mut request_headers = request.headers().clone();
+    let request_headers = request.headers().clone();
+
+    // Extract traceparent header early (Envoy should have added this)
+    let traceparent = request_headers
+        .get("traceparent")
+        .and_then(|h| h.to_str().ok())
+        .unwrap_or("00-00000000000000000000000000000000-0000000000000000-01")
+        .to_string();
+
```
adilhafeez: Don't assume and take `00-00000000000000000000000000000000-0000000000000000-01` as default traceparent. You should use type `let traceparent: Option<String>` to know if traceparent was present or not.

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -55,142 +69,52 @@ pub async fn router_chat(
...
-        },
+        Ok(result) => result,
         Err(err) => {
-            let err_msg = format!("Failed to determine route: {}", err);
-            let mut internal_error = Response::new(full(err_msg));
-            *internal_error.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
+            let mut internal_error = Response::new(full(err.message));
+            *internal_error.status_mut() = err.status_code;
             return Ok(internal_error);
         }
```
adilhafeez, line 107: should we use default route if couldn't determine the route?

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -55,142 +69,52 @@ pub async fn router_chat(
...
-            trace_parent.clone(),
-            usage_preferences,
-        )
-        .await
+    // Determine routing using the dedicated router_chat module
+    let routing_result = match router_chat_get_upstream_model(
+        router_service,
+        client_request,  // Pass the original request - router_chat will convert it
+        &request_headers,
+        trace_collector.clone(),
```
adilhafeez, line 96: can trace_collector become variable like access it using `self.trace_collector`?

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -55,142 +69,52 @@ pub async fn router_chat(
...
         "[ARCH_ROUTER] URL: {}, Resolved Model: {}",
         full_qualified_llm_provider_url, model_name
     );
 
+    // Extract trace_parent for upstream request headers
+    let trace_parent = request_headers
+        .iter()
+        .find(|(ty, _)| ty.as_str() == "traceparent")
+        .map(|(_, value)| value.to_str().unwrap_or_default().to_string());
+
```
adilhafeez: only insert traceparent if its present - default wouldn't add any value

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+///
+/// # Returns
+/// * `Ok(RoutingResult)` - Contains the selected model name and span ID
+/// * `Err(RoutingError)` - Contains error details and optional span ID
+pub async fn router_chat_get_upstream_model(
+    router_service: Arc<RouterService>,
+    client_request: ProviderRequestType,
+    request_headers: &hyper::HeaderMap,
+    trace_collector: Arc<TraceCollector>,
+    traceparent: &str,
```
adilhafeez, line 41: this should be optional

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+impl RoutingError {
+    pub fn internal_error(message: String) -> Self {
+        Self {
+            message,
+            status_code: StatusCode::INTERNAL_SERVER_ERROR
+        }
+    }
+}
+
+/// Determines the routing decision if
```
adilhafeez, line 31: truncated comment

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+            ));
+        }
+        Err(err) => {
+            warn!("Failed to convert request to ChatCompletionsRequest: {}", err);
+            return Err(RoutingError::internal_error(format!(
+                "Failed to convert request: {}",
+                err
+            )));
+        }
+    };
```
adilhafeez, line 73: this is probably client error and not server error as it was caused by invalid user request

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+        "[ARCH_ROUTER REQ]: {}",
+        &serde_json::to_string(&chat_request).unwrap()
+    );
+
+    // Extract trace_parent from headers
+    let trace_parent = request_headers
+        .iter()
+        .find(|(ty, _)| ty.as_str() == "traceparent")
+        .map(|(_, value)| value.to_str().unwrap_or_default().to_string());
+
```
adilhafeez, line 85: don't use `unwrap_or_default` - if traceparent doesn't exist don't assume default

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+        .as_ref()
+        .and_then(|s| serde_yaml::from_str(s).ok());
+
+    // Prepare log message with latest message from chat request
+    let latest_message_for_log = chat_request
+        .messages
+        .last()
+        .map_or("None".to_string(), |msg| {
+            msg.content.to_string().replace('\n', "\\n")
+        });
```
adilhafeez, line 103: I think you could use get_recent_message method to get latest message

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+    const MAX_MESSAGE_LENGTH: usize = 50;
+    let latest_message_for_log = if latest_message_for_log.chars().count() > MAX_MESSAGE_LENGTH {
+        let truncated: String = latest_message_for_log
+            .chars()
+            .take(MAX_MESSAGE_LENGTH)
+            .collect();
+        format!("{}...", truncated)
+    } else {
+        latest_message_for_log
+    };
```
adilhafeez, line 114: I think you've used truncate function in router.rs make extract it out in a function and use it here?

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+        latest_message_for_log
+    );
+
+    debug!("usage preferences from request: {:?}", usage_preferences);
+
+    // Capture start time for routing span
+    let routing_start_time = std::time::Instant::now();
+    let routing_start_system_time = std::time::SystemTime::now();
+
+    // Attempt to determine route using the router service
```
adilhafeez, line 129: why use Instant and SystemTime both here?

crates/brightstaff/src/handlers/router_chat.rs
```diff
@@ -0,0 +1,239 @@
...
+                routing_start_time,
+                routing_start_system_time,
+                attrs,
+            )
+            .await;
+
+            Err(RoutingError::internal_error(
+                format!("Failed to determine route: {}", err)
+            ))
+        }
```
adilhafeez, line 194: I posted another comment in router.rs which is should we fail the request in case of routing failure or just use default route

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -1,32 +1,180 @@
...
 
     /// Called when streaming encounters an error
     fn on_error(&mut self, _error: &str) {}
 }
 
-/// A no-op processor that just forwards chunks as-is
-pub struct PassthroughProcessor;
+/// A processor that tracks streaming metrics and finalizes the span
+pub struct PassthroughProcessor {
+    collector: Arc<TraceCollector>,
```
adilhafeez: it should now be called sth that conveys that this processor is storing metrics sth like StreamMetricsProcessor

crates/common/src/traces/collector.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// - `TRACE_FLUSH_INTERVAL_SECS` - Flush interval in seconds (default: 1)
+    /// - `OTEL_COLLECTOR_URL` - OTEL collector endpoint (default: http://localhost:9903/v1/traces)
+    pub fn new(enabled: Option<bool>) -> Self {
+        let flush_interval_secs = std::env::var("TRACE_FLUSH_INTERVAL_SECS")
+            .ok()
+            .and_then(|s| s.parse().ok())
+            .unwrap_or(1);
+
+        let otel_url = std::env::var("OTEL_COLLECTOR_URL")
+            .unwrap_or_else(|_| "http://localhost:9903/v1/traces".to_string());
```
adilhafeez, line 75: I already left a note at envoy template but you can directly send traces to otel collected don't need to proxy it through envoy

crates/common/src/traces/collector.rs
```diff
@@ -0,0 +1,274 @@
...
+                .entry(service_name)
+                .or_insert_with(VecDeque::new);
+
+            spans.push_back(span);
+        } else {
+            // Lock contention - skip recording this span
+            debug!("Skipped span recording due to lock contention");
+        }
+        // Flushing is handled by the periodic background flusher (see `start_background_flusher`).
+    }
```
adilhafeez, line 128: why store spans separately by service? why not just have one global queue for all spans?

crates/common/src/traces/collector.rs
```diff
@@ -0,0 +1,274 @@
...
+        match self.send_to_otel(resource_spans).await {
+            Ok(_) => {
+                debug!("Successfully flushed {} spans", total_spans);
+                Ok(())
+            }
+            Err(e) => {
+                warn!("Failed to send spans to OTEL collector: {:?}", e);
+                Err(e)
+            }
+        }
```
adilhafeez, line 177: just fyi - if http call fails here you will lose traces

arch/envoy.template.yaml
salmanap: plano(outbound) has routing and llm traces separately. So perhaps best to call it plano(outbound)

arch/envoy.template.yaml
salmanap: We do want to keep that for backwards compatibility - using 127.0.0.1.

crates/brightstaff/src/handlers/router.rs
salmanap: fixed

crates/brightstaff/src/handlers/router_chat.rs
salmanap: if the router fails - then we go to default model. We don't make strict guarantees. We go to default model.

crates/brightstaff/src/handlers/utils.rs
salmanap: done. "ObservableStreamProcessor"

crates/common/src/traces/collector.rs
salmanap: aggregated by service - so that spans are part of a single service.

crates/common/src/traces/collector.rs
salmanap: thanks for your comment.

Issue #614, opened 2025-11-08
Full support for responses API plus conversation api like functionality


ockhamlabs:
We currently use responses API along with conversational api for full context engineered ai agentic conversations and actions (,openai). Will be great to enable same drop-in replacement using any model and using memory /custom memory solution to provide same result

salmanap:
This is a well requested feature and we will need to use a short term cache or database to support that. Looking at options now

salmanap:
Shipped @ockhamlabs !

PR #631, opened 2025-12-15
enable state management for v1/responses

salmanap:
This PR completes the issue open: https://github.com/katanemo/archgw/issues/614 - where we manage the state of the conversation to offer a unified v1/responses API that can be used for any upstream LLM while the client is OpenAI

### DIFF
--- a/arch/arch_config_schema.yaml
+++ b/arch/arch_config_schema.yaml
@@ -331,7 +331,7 @@ properties:
       model:
         type: string
       additionalProperties: false
-  state_storage_v1_responses:
+  state_storage:
     type: object
     properties:
       type:
--- a/arch/tools/cli/utils.py
+++ b/arch/tools/cli/utils.py
@@ -161,6 +161,10 @@ def get_llm_provider_access_keys(arch_config_file):
             matches = re.findall(pattern, connection_string)
             for var in matches:
                 access_key_list.append(f"${var}")
+        else:
+            raise ValueError(
+                "Invalid connection string received in state_storage_v1_responses"
+            )
 
     return access_key_list
 
--- a/crates/brightstaff/src/state/memory.rs
+++ b/crates/brightstaff/src/state/memory.rs
@@ -47,14 +47,14 @@ impl StateStorage for MemoryConversationalStorage {
         match storage.get(response_id) {
             Some(state) => {
                 debug!(
-                    "[PLANO | BRIGHTSTAFF | MEMORY_STORAGE] RESP_ID:{} | Retrieved conversation state: input_items={}",
+                    "[PLANO | MEMORY_STORAGE | RESP_ID:{} | Retrieved conversation state: input_items={}",
                     response_id, state.input_items.len()
                 );
                 Ok(state.clone())
             }
             None => {
                 warn!(
-                    "[PLANO | BRIGHTSTAFF | MEMORY_STORAGE] RESP_ID:{} | Conversation state not found",
+                    "[PLANO_RESP_ID:{} | MEMORY_STORAGE | Conversation state not found",
                     response_id
                 );
                 Err(StateStorageError::NotFound(response_id.to_string()))
@@ -85,16 +85,16 @@ impl StateStorage for MemoryConversationalStorage {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use hermesllm::apis::openai_responses::{InputItem, InputMessage, MessageRole, InputContent};
+    use hermesllm::apis::openai_responses::{InputItem, InputMessage, MessageRole, InputContent, MessageContent};
 
     fn create_test_state(response_id: &str, num_messages: usize) -> OpenAIConversationState {
         let mut input_items = Vec::new();
         for i in 0..num_messages {
             input_items.push(InputItem::Message(InputMessage {
                 role: if i % 2 == 0 { MessageRole::User } else { MessageRole::Assistant },
-                content: vec![InputContent::InputText {
+                content: MessageContent::Items(vec![InputContent::InputText {
                     text: format!("Message {}", i),
-                }],
+                }]),
             }));
         }
 
@@ -222,9 +222,9 @@ mod tests {
         // Create current input with 1 message
         let current_input = vec![InputItem::Message(InputMessage {
             role: MessageRole::User,
-            content: vec![InputContent::InputText {
+            content: MessageContent::Items(vec![InputContent::InputText {
                 text: "New message".to_string(),
-            }],
+            }]),
         })];
 
         // Merge
@@ -244,24 +244,30 @@ mod tests {
         // Current input has message 2
         let current_input = vec![InputItem::Message(InputMessage {
             role: MessageRole::User,
-            content: vec![InputContent::InputText {
+            content: MessageContent::Items(vec![InputContent::InputText {
                 text: "Message 2".to_string(),
-            }],
+            }]),
         })];
 
         let merged = storage.merge(&prev_state, current_input);
 
         // Verify order: prev messages first, then current
-        let InputItem::Message(msg) = &merged[0];
-        match &msg.content[0] {
-            InputContent::InputText { text } => assert_eq!(text, "Message 0"),
-            _ => panic!("Expected InputText"),
+        let InputItem::Message(msg) = &merged[0] else { panic!("Expected Message") };
+        match &msg.content {
+            MessageContent::Items(items) => match &items[0] {
+                InputContent::InputText { text } => assert_eq!(text, "Message 0"),
+                _ => panic!("Expected InputText"),
+            },
+            _ => panic!("Expected MessageContent::Items"),
         }
 
-        let InputItem::Message(msg) = &merged[2];
-        match &msg.content[0] {
-            InputContent::InputText { text } => assert_eq!(text, "Message 2"),
-            _ => panic!("Expected InputText"),
+        let InputItem::Message(msg) = &merged[2] else { panic!("Expected Message") };
+        match &msg.content {
+            MessageContent::Items(items) => match &items[0] {
+                InputContent::InputText { text } => assert_eq!(text, "Message 2"),
+                _ => panic!("Expected InputText"),
+            },
+            _ => panic!("Expected MessageContent::Items"),
         }
     }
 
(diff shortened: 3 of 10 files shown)


arch/tools/cli/utils.py
```diff
@@ -148,6 +148,20 @@ def get_llm_provider_access_keys(arch_config_file):
...
+        connection_string = state_storage.get("connection_string")
+        if connection_string and isinstance(connection_string, str):
+            # Extract all $VAR and ${VAR} patterns from connection string
+            import re
+
+            # Match both $VAR and ${VAR} patterns
+            pattern = r"\$\{?([A-Z_][A-Z0-9_]*)\}?"
+            matches = re.findall(pattern, connection_string)
+            for var in matches:
+                access_key_list.append(f"${var}")
```
adilhafeez, line 163: else case you want to throw exception saying "invalid connection string received"

arch/arch_config_schema.yaml
```diff
@@ -331,6 +331,31 @@ properties:
       model:
         type: string
       additionalProperties: false
+  state_storage_v1_responses:
```
adilhafeez: I'd prefer a generic name but then I see that this is specific to v1/responses but then I also see that other providers could provide sth like this in future then state_storage_v1_responses will need an update. So think its better to keep it generic now,

crates/brightstaff/src/state/memory.rs
```diff
@@ -0,0 +1,584 @@
...
+        let storage = self.storage.read().await;
+
+        match storage.get(response_id) {
+            Some(state) => {
+                debug!(
+                    "[PLANO | BRIGHTSTAFF | MEMORY_STORAGE] RESP_ID:{} | Retrieved conversation state: input_items={}",
+                    response_id, state.input_items.len()
+                );
+                Ok(state.clone())
+            }
```
adilhafeez, line 54: you probably don't need clone here and you could move var

crates/brightstaff/src/state/postgresql.rs
```diff
@@ -0,0 +1,432 @@
...
+            .map_err(|e| {
+                StateStorageError::StorageError(format!("Failed to connect to database: {}", e))
+            })?;
+
+        // Spawn the connection to run in the background
+        tokio::spawn(async move {
+            if let Err(e) = connection.await {
+                warn!("Database connection error: {}", e);
+            }
+        });
```
adilhafeez, line 30: Why spawn? We could wait for a bit and fail/retry if connection fails. We could also do lazy init where we only create connection if there is a request to store/retrieve and close connection if there is no activity for some time.

crates/brightstaff/src/state/postgresql.rs
```diff
@@ -0,0 +1,432 @@
...
+
+#[async_trait]
+impl StateStorage for PostgreSQLConversationStorage {
+    async fn put(&self, state: OpenAIConversationState) -> Result<(), StateStorageError> {
+        self.ensure_ready().await?;
+
+        // Serialize input_items to JSONB
+        let input_items_json = serde_json::to_value(&state.input_items).map_err(|e| {
+            StateStorageError::StorageError(format!("Failed to serialize input_items: {}", e))
+        })?;
```
adilhafeez, line 86: you could store entire object rather than just data.input_items

crates/brightstaff/src/state/postgresql.rs
```diff
@@ -0,0 +1,432 @@
...
+                }
+
+                info!("Conversation state storage table verified");
+                Ok(())
+            })
+            .await?;
+
+        Ok(())
+    }
+}
```
adilhafeez, line 76: this whole test can be skipped if you initialize table during connection init

crates/hermesllm/src/transforms/response/to_openai.rs
```diff
@@ -80,8 +80,15 @@ impl TryFrom<ChatCompletionsResponse> for ResponsesAPIResponse {
             // Only add the message item if there's actual content (text, audio, or refusal)
             // Don't add empty message items when there are only tool calls
             if !content.is_empty() {
+                // Avoid double-prefixing: if ID already starts with "msg_", use as-is
+                let message_id = if resp.id.starts_with("msg_") {
+                    resp.id.clone()
+                } else {
+                    format!("msg_{}", resp.id)
+                };
```
adilhafeez: why do we need to prefix it with msg_ why can't we use id as it is?

crates/hermesllm/src/transforms/response/to_openai.rs
```diff
@@ -151,7 +158,12 @@ impl TryFrom<ChatCompletionsResponse> for ResponsesAPIResponse {
         };
 
         Ok(ResponsesAPIResponse {
-            id: resp.id,
+            // Generate proper resp_ prefixed ID if not already present
+            id: if resp.id.starts_with("resp_") {
+                resp.id
+            } else {
+                format!("resp_{}", uuid::Uuid::new_v4().to_string().replace("-", ""))
+            },
```
adilhafeez, line 170: same here why do we need to prefix id with resp_

crates/hermesllm/src/transforms/response/output_to_input.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/// Converts a Vec of OutputItems into InputItems for conversation continuation
+pub fn outputs_to_inputs(outputs: &[OutputItem]) -> Vec<InputItem> {
+    outputs
+        .iter()
+        .filter_map(output_item_to_input_item)
+        .collect()
+}
+
```
adilhafeez, line 83: this is a very weird name - can you think of coming up with a better name like maybe convert_responses_output_to_input_item or sth similar

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -83,9 +101,77 @@ pub async fn llm_chat(
...
     if client_request.remove_metadata_key("archgw_preference_config") {
-        debug!("Removed archgw_preference_config from metadata");
+        debug!("[PLANO_REQ_ID:{}] Removed archgw_preference_config from metadata", request_id);
+    }
+
+    // === v1/responses state management: Determine upstream API and combine input if needed ===
+    // Do this BEFORE routing since routing consumes the request
+    // Only process state if state_storage is configured
+    let mut should_manage_state = false;
+    if is_responses_api_client && state_storage.is_some() {
```
adilhafeez, line 111: why is it set to false I don't see it getting set to anything

arch/tools/cli/utils.py
salmanap: Fixed.

arch/arch_config_schema.yaml
salmanap: okay. fixed.

crates/brightstaff/src/state/memory.rs
salmanap: Unfortunately, not a trivial change. We will need Arc<OpenAIConversationState> and all implementations will be forced into that structure. Is there an easier way to do this in Rust? The cloned object that is returned gets discarded after the request terminates. So it is a short-lived object. Let me know how you would like to proceed.

crates/brightstaff/src/state/postgresql.rs
salmanap: This is actually necessary for [tokio-postgres](vscode-file://vscode-app/Applications/Visual%20Studio%20Code%202.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) architecture. The Connection future must be continuously polled to drive the I/O, which is why spawning is the standard pattern for this library. 

Separately, I don't want to manage the lifecycle of the database in brightstaff. Think that is a slippery slope (should we delete it too at some point? would we support that for other providers like Dyanmo?)

crates/brightstaff/src/state/postgresql.rs
salmanap: Please see comment above - I would prefer we do't manage the lifecycle of the table in brightstaff/state. This feels like mixing concerns on the create/update/delete. All we need is a connection string and make sure our DB is ready for us, else we fail.

crates/brightstaff/src/state/postgresql.rs
salmanap: agree - that is possible. But if we let each storage implemenation handle the downstream storage specifics, we can offer a more tailored experience to the storage technology. For SQL tables having some additional respone data would be helpful to debug and filter rows (such as based on time). Else the developer will have to write specific logic to work with our storage mechanism

crates/hermesllm/src/transforms/response/to_openai.rs
salmanap: This is important because when we convert a ChatCompletionsRequest object to ResponseAPIResponse object. We need to ensure that the upstream client always see resp_id. Else the client will error out. In the case of some LLM providers they don't send any resp_id so we generate an UUID and use that for storage and retrieval.

crates/hermesllm/src/transforms/response/to_openai.rs
salmanap: You caught a subtle bug format!("msg_{}}, resp.id) could produce msg_resp_ which would be wrong. Fixing that.

This is another ChatCompletionsResponse to ResponseAPIResponse conversion tricks. Each output item that is generated by an upstream LLM must be converted to an output item with a message_id because that is what the upstream responsesAPI client expects. So if some of the the LLM providers don't send us a msg_ then we need to build one using the resp.id

crates/hermesllm/src/transforms/response/output_to_input.rs
salmanap: agree. fixed.

crates/brightstaff/src/handlers/llm.rs
salmanap: it is being used later on, where we check if the upstream path is also v1/responses then we don't pull additional content.

crates/brightstaff/src/state/postgresql.rs
salmanap: table_verified: Arc<OnceCell<()>> - this will ensure that the table ready operation is lazy loaded, and once we know it will return (). This will handle multiple callers and manage concurrency for us. Fixed

adilhafeez (approved):

PR #632, opened 2025-12-15
fixed mixed inputs from openai v1/responses api

salmanap:


tests/e2e/test_openai_responses_api_client.py
adilhafeez, line 1: test with `ItemReference` use as well

PR #634, opened 2025-12-17
fixed reasoning failures

salmanap:
we should build the right reasoning object. Else the conversion can fail in a boundary case where the upstream LLM returns "none" and we can't map it to low/medium/high for v1/responses/api

### DIFF
--- a/arch/supervisord.conf
+++ b/arch/supervisord.conf
@@ -9,7 +9,7 @@ stdout_logfile_maxbytes=0
 stderr_logfile_maxbytes=0
 
 [program:envoy]
-command=/bin/sh -c "python -m cli.config_generator && envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && envoy -c /etc/envoy.env_sub.yaml --component-log-level wasm:debug --log-format '[%%Y-%%m-%%d %%T.%%e][%%l] %%v' 2>&1 | tee /var/log/envoy.log | while IFS= read -r line; do echo '[archgw_logs]' \"$line\"; done"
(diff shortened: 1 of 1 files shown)


crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
```diff
@@ -0,0 +1,241 @@
...
+                    if is_incomplete_json {
+                        // Incomplete JSON - buffer for retry with next chunk
+                        self.incomplete_event_buffer = sse_event.raw_line.as_bytes().to_vec();
+                        break;
+                    } else {
+                        // Other error (unsupported event type, validation error, etc.)
+                        // Skip this event and continue processing others
+                        continue;
+                    }
+                }
```
adilhafeez, line 75: What if there are more events afterwards? That might be valid?

adilhafeez (approved):
I think this code here may be hiding the underlying issue where it sees sse events but last event which breaks does not have line break but sse_iter::try_from will return that invalid evnet as event data line

For now it may be okay to take this change but this is where we need to fix where if the chunk doesn't end of line break then carry that buffer forward for the next chunk.

```rust
    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
        // Parse as text-based SSE format
        let s = std::str::from_utf8(bytes)?;
        let lines: Vec<String> = s.lines().map(|line| line.to_string()).collect();
        Ok(SseStreamIter::new(lines.into_iter()))
    }
```


https://github.com/katanemo/archgw/blob/main/crates/hermesllm/src/apis/streaming_shapes/sse.rs#L247-L252

crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
salmanap: Only the corrupted line is discarded. So additional lines with new events on the same payload will work okay, because of the continue. The challenge is if there is an incomplete json on a line, how do we really know if the additional bytes aren't part of the same event or truly part of a new event. I suppose you could do a regex search for proper json start/end. But I think a corrupt line is a corrupt line. Thoughts?

salmanap: Tracking improvement here: https://github.com/katanemo/archgw/issues/636

PR #637, opened 2025-12-18
plano orchestration using plano orchestration 4b model

adilhafeez:


crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -215,94 +206,160 @@ async fn handle_agent_chat(
...
+                .map_err(AgentFilterChainError::from);
+        }
+
+        // For intermediate agents, collect the full response and pass to next agent
+        debug!("Collecting response from intermediate agent: {}", agent_name);
+        let response_text = response_handler.collect_full_response(llm_response).await?;
+
+        // Create a new message with the agent's response as assistant message
+        // and add it to the conversation history
+        current_messages.push(OpenAIMessage {
```
salmanap, line 357: This will push a very specific shape. How will we support v1/messages v1/responses etc?

salmanap: Are we expecting that all agents upstream of Plano will be v1/chat/completions? That could be an okay assumption for this PR. But my sense is that we won't be able to necessarily enforce that because agents could be built using different API shapes like v1/responses and v1/messages or any API but with the right messages[] shape.

demos/use_cases/travel_agents/src/travel_agents/api.py
salmanap, line 1: I think we should avoid custom types in our code - it will be more learning for the developer. we should simple request.body['messages']. Else developers will have to open these shapes up and learn about them. I would prefer our demos to be dirt simple to understand, and the only concepts they are learning are about Plano

demos/use_cases/travel_agents/src/travel_agents/__init__.py
salmanap, line 1: @Spherrrical we should rename the demo folder from travel_booking to "agent_orchestration(travel)" - this is so that developers are clearer about the use case we are helping them build. vs a specific travel scenario

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -215,94 +206,160 @@ async fn handle_agent_chat(
...
+        trace_collector.record_span(operation_component::AGENT, span);
+
+        // If this is the last agent, return the streaming response
+        if is_last_agent {
+            info!("Completed agent chain, returning response from last agent: {}", agent_name);
+            return response_handler
+                .create_streaming_response(llm_response)
+                .await
+                .map_err(AgentFilterChainError::from);
+        }
```
salmanap, line 340: shouldn't we record traces for when the final agent completes its work?

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -215,94 +206,160 @@ async fn handle_agent_chat(
...
+
+        let span = span_builder.build();
+        trace_collector.record_span(operation_component::AGENT, span);
+
+        // If this is the last agent, return the streaming response
+        if is_last_agent {
+            info!("Completed agent chain, returning response from last agent: {}", agent_name);
+            return response_handler
+                .create_streaming_response(llm_response)
```
salmanap, line 337: This creates a hard depedency on every agent to give us a streaming response. Shouldn't we set "stream":true as default and honor some config from the developer where we can instantiate different types of calls. 

Also, I think we should debate if we should send traffic to an agent via the deprecated v1/chat/completions API used by only OpenAI or just /invoke where we would send the messages[] and perhaps a simplified version of ChatCompletionsRequest called AgentRequest which would include things like stream:true and other necessary fields (technically, we could detect the response type like stream/no-stream from the headers and make sure we do the right thing on our side)

This is how developers call an agent in LanghChain

```
# Create agent (example with model and tools)
from langchain import createAgent
# ... agent setup ...

# Single user message
result = agent.invoke({
    "messages": [HumanMessage(content="What's the weather in SF?")]
})
```

This is how it works in the openai sdk

```
    agent = Agent(
        name="Knowledgable GPT-5 Assistant",
        instructions="You're a knowledgable assistant. You always provide an interesting answer.",
        model="gpt-5",
        model_settings=ModelSettings(
            reasoning=Reasoning(effort="minimal"),  # "minimal", "low", "medium", "high"
            verbosity="low",  # "low", "medium", "high"
        ),
    )
    result = await Runner.run(agent, "Tell me something about recursion in programming.")
    print(result.final_output)
```

Of course these are SDK implementations, and we need to decide on the most idiomatic version of this from plano. But something for us to consider as we expose this programming model for agents via Plano.

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: That is good observation. We took two assumptions here 1) mcp agents will exclusively use [Message] which is openai's ChatMessage format and 2) to call agent I simplified upstream to use v1/chat/completions. But for upstream we could use any of the supported format which is v1/responses, v1/messages or chat/completions because we have already done the translation work. Let me review the work and make appropriate changes.

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: > we should debate if we should send traffic to an agent via the deprecated v1/chat/completions
we shouldn't send traffic via v1/chat/completions as mentioned in earlier comment we should send use same format as used/dictated by agents definition in arch_config.yaml. Right now we don't specify the path and assume /v1/chat/completions but a developer could define agent like this,

```
agents:
  - id: weather_agent
    url: http://host.docker.internal:10510/v1/chat/completions
  - id: flight_agent
    url: http://host.docker.internal:10520/v1/responses
  - id: currency_agent
    url: http://host.docker.internal:10530/v1/messages
```

Those code samples you pasted above are very specific implementation of agents. We are not playing at that level. We are step above where we work off from filter url or an agent url. We (plano) understands protocol and are able to translate between different providers and would help connect different agents together.

demos/use_cases/travel_agents/src/travel_agents/api.py
adilhafeez: That is good comment. I tired to do that with mcp filter demo. Let me get this fixed here.

demos/use_cases/travel_agents/src/travel_agents/__init__.py
adilhafeez: We should avoid introducing brackets in the path, as this is not considered standard practice and is generally discouraged. My suggestion is to use keep travel_agents as this does convey multi-agent concept. And we anyway have to restructure demos. Let's do the overhaul in structuring the demos in followup PR.

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: Envoy is recording this trace. Look at `invoke_agent` implementation,

```
...
        // Set traceparent header to make the egress span a child of the agent span
        if !trace_id.is_empty() && !agent_span_id.is_empty() {
            let trace_parent = format!("00-{}-{}-01", trace_id, agent_span_id);
            agent_headers.remove(TRACE_PARENT_HEADER);
            agent_headers.insert(
                TRACE_PARENT_HEADER,
                hyper::header::HeaderValue::from_str(&trace_parent).unwrap(),
            );
        }
...
```

demos/use_cases/travel_agents/src/travel_agents/__init__.py
salmanap: Fair enough

crates/brightstaff/src/handlers/agent_chat_completions.rs
adilhafeez: > streaming 

This `create_streaming_response` method doesn't depend on streaming/non-streaming response that is set in the request body it's a generic handler that forwards data as it receives.

salmanap (approved):

PR #639, opened 2025-12-20
Update docs and README to Plano

salmanap:


README.md
adilhafeez, line 1: Our main readme.md should be minimal so the source of truth is docs.planoni.dev and we won't have inconsistencies to manage. It should list highlevel details and have pointers to hosted docs.

adilhafeez (commented):
- we should disable dark mode as it creates inconsistent views for pages that have images for example see `intro_to_plano` page
- our main readme.md should be minimal, only highlight key features and should have outbound links to hosted docs so that we will never have inconsistent views
- on side panel we should make main topic clickable and like "getting started", "concepts", "guides" should be clickable and lead to first page in that section
- the main homepage should point to overview rather than creating a separate page
- in the overview page we shouldn't list subsection as it creates complexity rather than keep thigns simple
- in "intro to plano" page "on this page" section is missing
- "next steps" in left nav is un-nessesary and is out of place
- "quick start" we should lead with simplest config and one that is easy to explain which is llm proxy, then explain llm routing, and lastly talk about agent orchestration. And skip completely prompt targets for now.
- "quick start" - think this page is quite detailed I'm not sure what but it can be simplified and borken down into our core offerings like "quick start for llm routing", "quick start for agent orchestration" ...
- "Access Logging" - we have stopped mapping logging path to local instead we use docker native command to pull access logs - I will update the docs. For example this command `tail -F ~/archgw_logs/access_*.log` is invalid
- "Deployment" we should talk about how to deploy to aws/gcloud using ecs or similar tech

PR #655, opened 2025-12-25
Introduce signals change

salmanap:


crates/brightstaff/src/handlers/llm.rs
```diff
@@ -287,13 +290,18 @@ pub async fn llm_chat(
...
         operation_component::LLM,
         llm_span,
         request_start_time,
     );
 
+    // Add messages for signal analysis if available
+    if !messages_for_signals.is_empty() {
+        base_processor = base_processor.with_messages(messages_for_signals);
+    }
+
```
adilhafeez: This is misuse of builder pattern - with `builder.with_..` you want to add additional optionals which are needed to construct final instance using `builder.build()` but that is not the case here.

I think you probably need base_processor.set_messages(...) method here

adilhafeez: its probably better to place it when constructing ObservableStreamProcessor at line 293

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -133,6 +143,94 @@ impl StreamProcessor for ObservableStreamProcessor {
...
+                    },
+                });
+
+                self.span.attributes.push(Attribute {
+                    key: signal_constants::REPAIR_RATIO.to_string(),
+                    value: AttributeValue {
+                        string_value: Some(format!("{:.3}", report.follow_up.repair_ratio)),
+                    },
+                });
+            }
```
adilhafeez, line 170: repair_count/repair_ratio seems quite specific to a market - I think we should use generic terms

crates/brightstaff/src/signals/analyzer.rs
```diff
@@ -0,0 +1,2592 @@
...
+                } else if Self::should_use_fuzzy_matching(pattern)
+                    && norm_msg.fuzzy_contains_phrase(pattern, self.fuzzy_threshold)
+                {
+                    repair_count += 1;
+                    repair_phrases.push(format!("Turn {}: '{}' (fuzzy)", i + 1, pattern));
+                    found_in_turn = true;
+                    break;
+                }
+            }
+
```
adilhafeez: Runtime for this algorithm is a concern because there is lots of text matching happening and it increases linearly with number of tokens on the input. Let's start logging runtime for this function to track its runtime.

crates/brightstaff/src/signals/analyzer.rs
```diff
@@ -0,0 +1,2592 @@
...
+        let overall_quality = self.assess_overall_quality(
+            &turn_count,
+            &follow_up,
+            &frustration,
+            &repetition,
+            &positive_feedback,
+            &escalation,
+        );
+
```
adilhafeez: This is definitely a good start but it is opinionated - and we may need different numbers/values on what good vs bad mean depending on user case study.

crates/brightstaff/src/signals/analyzer.rs
```diff
@@ -0,0 +1,2592 @@
...
+/// Main analyzer that computes all signals from a message array
+pub struct SignalAnalyzer {
+    /// Baseline expected turns for normal interactions
+    baseline_turns: usize,
+    /// Threshold for fuzzy pattern matching (0.0-1.0)
+    fuzzy_threshold: f64,
+}
+
+impl SignalAnalyzer {
```
adilhafeez: This should be an trait for example (design to interface pattern),

```
trait SignalAnalyzer {
  fn analyze(&self, messages: &[Message]) -> SignalReport;
}

pub struct TextBaedSignalAnalyzer {

}

impl SignalAnalyzer for TextBaedSignalAnalyzer {
  ...
}
```

This will help us provide different implementation in future easier

crates/brightstaff/src/signals/analyzer.rs
salmanap: This happens async - this will not be a blocking call because its in the thread that handles the flushing of traces. I agree, however there is a potential a soft thread could hang and then we wouldn't know how to terminate that and recover neatly. But it won't be impacting end users.

crates/brightstaff/src/signals/analyzer.rs
salmanap: Agree - we can enable knobs if this features takes off so that developers can define their own thresholds for signals

crates/brightstaff/src/handlers/utils.rs
salmanap: @nehcgs to have a deeper look at the signals taxonomy

crates/brightstaff/src/signals/analyzer.rs
salmanap: Added bounds. Although I think there could be some run away computational scenarios that we should address carefully.

crates/brightstaff/src/handlers/llm.rs
salmanap: fixed.

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -292,6 +295,7 @@ pub async fn llm_chat(
         operation_component::LLM,
         llm_span,
         request_start_time,
+        (!messages_for_signals.is_empty()).then_some(messages_for_signals),
     );
```
adilhafeez: I think it can be simplified by just replacing this entire statement with `Some(messages_for_signals)`

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -133,6 +139,94 @@ impl StreamProcessor for ObservableStreamProcessor {
             }
         }
 
+        // Analyze signals if messages are available and add to span attributes
+        if let Some(ref messages) = self.messages {
+            let analyzer = TextBasedSignalAnalyzer::new();
+            let report = analyzer.analyze(messages);
```
adilhafeez: To ensure that trait is used and you don't accidentally use any non-contract methods you'd have to create trait var. Like this code change,

```
            let analyzer: SignalAnalyzer = TextBasedSignalAnalyzer::new();
```

PR #676, opened 2026-01-06
Musa/demo fix

Spherrrical:
Fixing the demo due to orchestration issues and unnecessary changes


salmanap (commented):
We should strip down the fix to the demo to the most absolute fix: does multi-agent orchestration work. Everything else is bonus and can come later: like the create_sse_event and other helper functions. Please consider fixing.

Spherrrical:
> We should strip down the fix to the demo to the most absolute fix: does multi-agent orchestration work. Everything else is bonus and can come later: like the create_sse_event and other helper functions. Please consider fixing.

image

I've went ahead and removed the SSE chunk render. Context is provided through the agent, and if nothing is found it goes through a closest alternative. Ex. shown here Seattle to Lahore, but via Doha.

PR #694, opened 2026-01-16
don't include internal models during random model selection

adilhafeez:


salmanap (changes requested):
actively making changes in this path with support for wildcard model names. I can bring this code change if urgent. But I would hold off on this

PR #696, opened 2026-01-16
Adding support for wildcard models in the model_providers config

salmanap:
We should be able to support the following config so that developers don't have to restart plano everytime they want to try/use a new model. 

```version: v0.3.0

listeners:
  # Model listener for direct LLM access
  - type: model
    name: llms
    address: 0.0.0.0
    port: 12000

model_providers:
  # OpenAI - support all models via wildcard
  - model: openai/*
    access_key: $OPENAI_API_KEY

  # Anthropic - support all Claude models
  - model: anthropic/*
    access_key: $ANTHROPIC_API_KEY

  - model: xai/*
    access_key: $GROK_API_KEY


  # Custom internal LLM provider
  # Note: Requires base_url and provider_interface for unknown providers
  - model: ollama/*
    base_url: https://llm.internal.company.com
```

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -19,9 +19,8 @@ jobs:
       # Build and run the Docker container to generate the documentation
       - name: Build documentation using Docker
         run: |
-          cd ./docs
-          chmod +x build_docs.sh
-          ./build_docs.sh
+          chmod +x docs/build_docs.sh
+          sh docs/build_docs.sh
 
       - name: Copy CNAME to HTML Build Directory
         run: cp docs/CNAME docs/build/html/CNAME


cli/planoai/config_generator.py
```diff
@@ -187,11 +187,21 @@ def validate_and_render_schema():
...
-            if model_name in model_name_keys:
+
+            # Check if this is a wildcard model (provider/*)
+            is_wildcard = False
+            if "/" in model_name:
+                model_name_tokens = model_name.split("/")
+                if len(model_name_tokens) >= 2 and model_name_tokens[-1] == "*":
+                    is_wildcard = True
+
+            if model_name in model_name_keys and not is_wildcard:
```
adilhafeez, line 198: what if someone types `model = "*"`, we don't support that it seems like and we should just throw error

cli/planoai/config_generator.py
```diff
@@ -187,11 +187,21 @@ def validate_and_render_schema():
...
                 raise Exception(
                     f"Duplicate model name {model_name}, please provide unique model name for each model_provider"
                 )
```
adilhafeez, line 201: How would this work with this scenario where model_ids are duplicate let's say both openai and ollama has gpt-4o. This will result in inconsistent behavior where user could just request `model = gpt-4o`, which provider would we forward the request to?


```
model_providers:
  - model: openai/*
  - model: ollama/*
``` 

I do think better compromise would be if there is duplicate model we fail unless developer request model with provider name for example in the above scenario if user requested `model = openai/gpt-4o` the request should succeed.

config/supervisord.conf
```diff
@@ -19,7 +19,7 @@ command=/bin/sh -c "\
     uv run python -m planoai.config_generator && \
     envsubst < /etc/envoy/envoy.yaml > /etc/envoy.env_sub.yaml && \
     envoy -c /etc/envoy.env_sub.yaml \
-          --component-log-level wasm:info \
+          --component-log-level wasm:debug \
```
adilhafeez:
```suggestion
          --component-log-level wasm:info \
```

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -123,6 +124,27 @@ pub async fn llm_chat(
...
+    if llm_providers.read().await.get(&resolved_model).is_none() {
+        let err_msg = format!(
+            "Model '{}' not found in configured providers",
+            resolved_model
+        );
+        warn!("[PLANO_REQ_ID:{}] | FAILURE | {}", request_id, err_msg);
+        let mut bad_request = Response::new(full(err_msg));
+        *bad_request.status_mut() = StatusCode::BAD_REQUEST;
+        return Ok(bad_request);
+    }
```
adilhafeez, line 138: we should define a consistent message - maybe not for this PR but we should respond consistently for example using following format,

```
HTTP/1.1 400 Bad Request
Content-Type: application/json
Content-Length: xx

{
  "error": "Bad request",
  "message": "error_message",
}
```

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -132,7 +154,9 @@ pub async fn llm_chat(
     // Extract messages for signal analysis (clone before moving client_request)
     let messages_for_signals = client_request.get_messages();
 
-    client_request.set_model(resolved_model.clone());
+    // Set the model to just the model name (without provider prefix)
+    // This ensures upstream receives "gpt-4" not "openai/gpt-4"
+    client_request.set_model(model_name_only.clone());
```
adilhafeez, line 159: repeated comment - pls remove

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -240,11 +264,20 @@ pub async fn llm_chat(
...
+    // Determine final model to use
+    // Router returns "none" as a sentinel value when it doesn't select a specific model
+    let router_selected_model = routing_result.model_name;
+    let model_name = if router_selected_model != "none" {
+        // Router selected a specific model via routing preferences
+        router_selected_model
+    } else {
+        // Router returned "none" sentinel, use validated resolved_model from request
+        resolved_model.clone()
+    };
```
adilhafeez, line 276: I think this should be `model_name_only`

crates/hermesllm/src/providers/id.rs
```diff
@@ -23,31 +53,70 @@ pub enum ProviderId {
...
+            ProviderId::Anthropic => "anthropic",
+            ProviderId::Mistral => "mistralai",
+            ProviderId::Deepseek => "deepseek",
+            ProviderId::Groq => "groq",
+            ProviderId::XAI => "x-ai",
+            ProviderId::Moonshotai => "moonshotai",
+            ProviderId::Zhipu => "z-ai",
+            ProviderId::Qwen => "qwen",
+            _ => return Vec::new(),
+        };
```
adilhafeez, line 96: we shouldn't return emtpy - rather we should fail here if invalid/unsupported provider is requested

crates/llm_gateway/src/stream_context.rs
```diff
@@ -128,16 +129,40 @@ impl StreamContext {
...
+                            self.request_identifier(),
+                            err
+                        );
+                        return Err(err);
+                    }
+                }
+            }
+        };
+
+        self.llm_provider = Some(provider);
```
adilhafeez, line 166: I think we should just make it a static function and return selected_llm for example line 794 should look like this,

```
self.llm_provider = select_llm_provider()?;
```

docs/source/concepts/llm_providers/supported_providers.rst
```diff
@@ -26,7 +26,7 @@ All providers are configured in the ``llm_providers`` section of your ``plano_co
 
 **Common Configuration Fields:**
 
-- ``model``: Provider prefix and model name (format: ``provider/model-name``)
+- ``model``: Provider prefix and model name (format: ``provider/model-name`` or ``provider/*`` for wildcard expansion)
 - ``access_key``: API key for authentication (supports environment variables)
```
adilhafeez, line 30: we should link to current static list here so developers know what models will be included

cli/planoai/config_generator.py
salmanap: This will be captured in the lines above. 

```           model_name_tokens = model_name.split("/")
            if len(model_name_tokens) < 2:
                raise Exception(
                    f"Invalid model name {model_name}. Please provide model name in the format <provider>/<model_id> or <provider>/* for wildcards."
                )```

cli/planoai/config_generator.py
salmanap: In the case of ollama/* which is not a first-class provider the client will have to offer the provider_id/model_id combination. Else we will send traffic to one of the first-class providers if the model name is present in that set. Note that none of the models from each first-class provider have any overlap

config/supervisord.conf
salmanap: fixed

crates/brightstaff/src/handlers/llm.rs
salmanap: I agree with this 100% - we need a proper error message hygiene through out the code base. Issue opened here. https://github.com/katanemo/plano/issues/706

crates/brightstaff/src/handlers/llm.rs
salmanap: fixed.

crates/brightstaff/src/handlers/llm.rs
salmanap: model_name_only is used for Arch-Router - the resolved_model is what's used for upstream calls to the LLM. I think this actually reveals a critical refactor in this brittle code path.

crates/hermesllm/src/providers/id.rs
salmanap: This is used in the pub fn models(&self) method that returns models for first-class providers. We shouldn't throw an error if the model provider is not a first-class provider. Instead, we return an empty list for custom-providers which wouldn't break the models/list API.

crates/llm_gateway/src/stream_context.rs
salmanap: Agree with the change. I'll make this a separate PR because there are some borrowing issues that I would have to resolve and run a whole bunch of tests on.

docs/source/concepts/llm_providers/supported_providers.rst
salmanap: great call out. Fixed

PR #705, opened 2026-01-27
[Feat]: add AWS STS assume-role and SigV4 signing for Bedrock

### LINKED ISSUE #690: Support AWS IAM auth for AWS Bedrock
Specifically, add support for role-based authentication, where Plano authenticates the Bedrock via an AWS IAM Role.  For more details start here: https://docs.aws.amazon.com/bedrock/latest/userguide/security-iam.html

Note, we don't want to package the RUST SDK with Plano. It would now allow us to control failover to other providers neatly, not follow a consistent retry mechanism and other scenarios where we want to manage the binary payload using Envoy WASM ourselves. We are okay to bring in select Rust crates as utilities to help us as needed. For example, today we use aws-smithy-eventstream = "0.60" to help manage the binary frames in which Bedrock sends back LLM response streams.

Out of scope: long-living sessions keys. Only support for IAM role-based authentication.

samadpls:
Fixes #690 
- Added support for AWS credentials configuration in the schema and configuration files.
- Introduced functions to retrieve AWS credentials from the environment and sign requests using AWS Signature Version 4.
- Updated the StreamContext and FilterContext to handle AWS credentials and integrate with Bedrock services.
- Added new error handling for AWS-related operations.
- Updated Cargo.toml files to include necessary AWS dependencies and features.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,115 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+Plano is an AI-native proxy server and data plane for agentic applications, built on Envoy proxy. It centralizes agent orchestration, LLM routing, observability, and safety guardrails as an out-of-process dataplane.
+
+## Build & Test Commands
+
+### Rust (crates/)
+
+```bash
+# Build WASM plugins (must target wasm32-wasip1)
+cd crates && cargo build --release --target=wasm32-wasip1 -p llm_gateway -p prompt_gateway
+
+# Build brightstaff binary (native target)
+cd crates && cargo build --release -p brightstaff
+
+# Run unit tests
+cd crates && cargo test --lib
+
+# Format check
+cd crates && cargo fmt --all -- --check
+
+# Lint
+cd crates && cargo clippy --locked --all-targets --all-features -- -D warnings
+```
+
+### Python CLI (cli/)
+
+```bash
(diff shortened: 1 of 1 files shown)


crates/llm_gateway/src/stream_context.rs
```diff
@@ -869,35 +1145,174 @@ impl HttpContext for StreamContext {
...
+                        None,
+                    );
+                    return Action::Pause;
+                };
 
-        let body_bytes = match self.get_http_request_body(0, body_size) {
-            Some(body_bytes) => body_bytes,
-            None => {
+                if let Err(e) =
+                    self.sign_bedrock_request_with_iam_and_body(&iam_auth_clone, &body_for_signing)
```
san81: This could be too early to sign! Check out here https://github.com/katanemo/plano/blob/main/crates/hermesllm/README.md#architecture

Plano uses `hermesllm` to generalize the multi provider request. As it could possibly modify the request, signing here could lead to signature mismatch for Bedrock

crates/llm_gateway/src/stream_context.rs
```diff
@@ -99,6 +125,135 @@ impl StreamContext {
             .cloned()
             .unwrap_or_else(|| "NO_REQUEST_ID".to_string())
     }
+
+    fn get_cached_credentials(&self, _role_arn: &str) -> Option<AwsCredentials> {
+        let cached = self.cached_sts_credentials.borrow();
```
san81: cached credentials should be per `_role_arn`?

Issue #707, opened 2026-01-27
Custom trace attributes from request headers


Spherrrical:
We want to enrich OTEL spans with business identifiers (e.g., workspace.id, tenant.id, user.id) so developers can filter and debug traces by the context that matters to them. Today, traces include request path and system metadata, but not application-specific identifiers. This makes it hard to troubleshoot per-tenant, per-workspace, or per-user issues.

aquacommander:
> We want to enrich OTEL spans with business identifiers (e.g., workspace.id, tenant.id, user.id) so developers can filter and debug traces by the context that matters to them. Today, traces include request path and system metadata, but not application-specific identifiers. This makes it hard to troubleshoot per-tenant, per-workspace, or per-user issues.

I added first-class support for enriching OTEL spans with business identifiers such as tenant.id, workspace.id, and user.id.
This is done via a new tracing.span_attribute_headers config that maps incoming HTTP headers to OTEL span attribute keys.
On each request, the headers are read once, matched case-insensitively, and converted into a set of business attributes.
These attributes are then propagated through the entire request lifecycle and attached to all relevant spans, including routing, LLM execution, agent selection, and filter/MCP execution.
Empty or missing headers are safely ignored, ensuring no noisy or incorrect span data.
With this in place, developers can now reliably filter and debug traces per tenant, workspace, or user in tools like Jaeger, Tempo, or Datadog.

PR #708, opened 2026-01-27
add Custom Trace Attributes to extend observability

Spherrrical:
Introduces Custom Trace Attributes to allow developers to extend business logic!

Today, traces include request path and system metadata, but not application-specific identifiers. This makes it hard to troubleshoot per-tenant, per-workspace, or per-user issues.

#707

### DIFF
--- a/crates/brightstaff/src/tracing/custom_attributes.rs
+++ b/crates/brightstaff/src/tracing/custom_attributes.rs
@@ -4,48 +4,6 @@ use common::configuration::SpanAttributes;
 use common::traces::SpanBuilder;
 use hyper::header::HeaderMap;
 
-pub fn extract_custom_trace_attributes(
-    headers: &HeaderMap,
-    span_attribute_header_prefixes: Option<&[String]>,
-) -> HashMap<String, String> {
-    let mut attributes = HashMap::new();
-    let Some(span_attribute_header_prefixes) = span_attribute_header_prefixes else {
-        return attributes;
-    };
-    if span_attribute_header_prefixes.is_empty() {
-        return attributes;
-    }
-
-    for (name, value) in headers.iter() {
-        let header_name = name.as_str();
-        let mut matched_prefix: Option<&str> = None;
-        for prefix in span_attribute_header_prefixes {
-            if header_name.starts_with(prefix) {
-                matched_prefix = Some(prefix.as_str());
-                break;
-            }
-        }
-        let Some(prefix) = matched_prefix else {
-            continue;
-        };
-
-        let raw_value = match value.to_str().ok() {
-            Some(value) => value.trim(),
-            None => continue,
-        };
-
-        let suffix = header_name.strip_prefix(prefix).unwrap_or("");
-        let suffix_key = suffix.trim_start_matches('-').replace('-', ".");
-        if suffix_key.is_empty() {
-            continue;
-        }
-
-        attributes.insert(suffix_key, raw_value.to_string());
-    }
-
-    attributes
-}
-
 pub fn collect_custom_trace_attributes(
     headers: &HeaderMap,
     span_attributes: Option<&SpanAttributes>,
@@ -61,10 +19,36 @@ pub fn collect_custom_trace_attributes(
         }
     }
 
-    attributes.extend(extract_custom_trace_attributes(
-        headers,
-        span_attributes.header_prefixes.as_deref(),
-    ));
+    let Some(header_prefixes) = span_attributes.header_prefixes.as_deref() else {
+        return attributes;
+    };
+    if header_prefixes.is_empty() {
+        return attributes;
+    }
+
+    for (name, value) in headers.iter() {
+        let header_name = name.as_str();
+        let matched_prefix = header_prefixes
+            .iter()
+            .find(|prefix| header_name.starts_with(prefix.as_str()))
+            .map(String::as_str);
+        let Some(prefix) = matched_prefix else {
+            continue;
+        };
+
+        let Some(raw_value) = value.to_str().ok().map(str::trim) else {
+            continue;
+        };
+
+        let suffix = header_name.strip_prefix(prefix).unwrap_or("");
+        let suffix_key = suffix.trim_start_matches('-').replace('-', ".");
+        if suffix_key.is_empty() {
+            continue;
+        }
+
+        attributes.insert(suffix_key, raw_value.to_string());
+    }
+
     attributes
 }
 
@@ -80,8 +64,10 @@ pub fn append_span_attributes(
 
 #[cfg(test)]
 mod tests {
-    use super::extract_custom_trace_attributes;
+    use super::collect_custom_trace_attributes;
+    use common::configuration::SpanAttributes;
     use hyper::header::{HeaderMap, HeaderValue};
+    use std::collections::HashMap;
 
     #[test]
     fn extracts_headers_by_prefix() {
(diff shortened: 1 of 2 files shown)


Spherrrical:
Implemented prefix-based trace attribute extraction and removed the old per-header schema. The gateway now reads tracing.custom_attribute_prefixes and captures any matching request headers as OTEL attributes, trimming the prefix so x-katanemo-workspace-id becomes workspace.id, with values recorded as strings. Updated config/schema and demo files to use the new prefix model, added a prefix-mismatch example in test.rest, and updated unit tests to match the new behavior.

image


To enable the custom prefix:

```yaml
tracing:
  random_sampling: 100
  custom_attribute_prefixes:
    - x-katanemo-
```

Headers must contain `x-[identifier]` for attribute to pass through OTEL by configuration.

config/arch_config_schema.yaml
```diff
@@ -382,7 +382,10 @@ properties:
         type: integer
       trace_arch_internal:
         type: boolean
-      additionalProperties: false
+      custom_attribute_prefixes:
```
salmanap: we should probably call this span_attribute_header_perfixes

crates/brightstaff/src/tracing/custom_attributes.rs
```diff
@@ -0,0 +1,68 @@
...
+        let Some(prefix) = matched_prefix else {
+            continue;
+        };
+
+        let raw_value = match value.to_str().ok() {
+            Some(value) => value.trim(),
+            None => continue,
+        };
+
+        let suffix = header_name.strip_prefix(prefix).unwrap_or("");
```
salmanap, line 43: There are some semantics here that I don't think we've made explicit to the developer that we will transform x-user-id to user.id in the traces. We should add a comment here and add a section in the docs to talk about this.

config/plano_config_schema.yaml
```diff
@@ -382,7 +382,10 @@ properties:
         type: integer
       trace_arch_internal:
         type: boolean
-      additionalProperties: false
```
adilhafeez, line 393: we still need `additionalProperties: false` here

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -176,6 +182,13 @@ async fn handle_agent_chat(
 
         headers
     };
+    let custom_attrs = extract_custom_trace_attributes(
+        &request_headers,
+        tracing_config
+            .as_ref()
+            .as_ref()
+            .and_then(|tracing| tracing.span_attribute_header_prefixes.as_deref()),
+    );
```
adilhafeez: should go other way where only look at request headers if custom tracing attributes are set - this will optimize the run time

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -359,6 +375,9 @@ async fn handle_agent_chat(
                 "duration_ms",
                 format!("{:.2}", agent_elapsed.as_secs_f64() * 1000.0),
             );
+        for (key, value) in &custom_attrs {
+            span_builder = span_builder.with_attribute(key, value);
+        }
```
adilhafeez: why is it repeating - I saw another instance of this code at line 285

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -26,25 +26,35 @@ use crate::state::response_state_processor::ResponsesStateProcessor;
...
+#[allow(clippy::too_many_arguments)]
 pub async fn llm_chat(
     request: Request<hyper::body::Incoming>,
     router_service: Arc<RouterService>,
     full_qualified_llm_provider_url: String,
     model_aliases: Arc<Option<HashMap<String, ModelAlias>>>,
     llm_providers: Arc<RwLock<LlmProviders>>,
     trace_collector: Arc<TraceCollector>,
+    tracing_config: Arc<Option<Tracing>>, // ! right here
     state_storage: Option<Arc<dyn StateStorage>>,
```
adilhafeez: these comments are of no value - remove them

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -26,25 +26,35 @@ use crate::state::response_state_processor::ResponsesStateProcessor;
...
     let request_path = request.uri().path().to_string();
     let request_headers = request.headers().clone();
+    let custom_attrs = extract_custom_trace_attributes(
+        &request_headers,
+        tracing_config
+            .as_ref()
+            .as_ref()
+            .and_then(|tracing| tracing.span_attribute_header_prefixes.as_deref()),
+    );
     let request_id: String = match request_headers
```
adilhafeez: go from other way that is if span_attributes are defined only then extract the attributes from header

crates/brightstaff/src/main.rs
```diff
@@ -112,6 +112,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     ));
 
     let model_aliases = Arc::new(arch_config.model_aliases.clone());
+    let tracing_config = Arc::new(arch_config.tracing.clone());
```
adilhafeez: don't send entire object may just send custom tracing block

adilhafeez (changes requested):
Thanks for putting this togehter. I do have some suggestions,


*Name can be a bit better* while I don't have an answer here but it should be sth like. Notice the `static` block too,

```yaml
tracing:
  ...
  span_attributes: # or trace_attributes or just attributes
    header_prefixes: # or from_header
      - x-katanemo-
    static:
      environment: production
      service.version: "1.0.0"
      deployment.region: us-west-2
  
```


*Placement of attributes in span* - I see that in the screenshot attached to plano(llm) span. For a single request will we see these attributes in all spans that are created in a trace? I think these attributes should be part of just the root trace.

crates/brightstaff/src/handlers/agent_chat_completions.rs
```diff
@@ -243,6 +257,13 @@ async fn handle_agent_chat(
         (String::new(), None)
     };
 
+    let apply_custom_attrs = |mut builder: SpanBuilder| {
```
salmanap: I don't think you want a lambda function in the file. Feels a bit misplaced. I think we should ask ourselves, whats the canonical way to append attributes after a span has been built. A simple for loop feels okay. Although I might move some of that functionality in commons

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -43,18 +42,31 @@ pub async fn llm_chat(
...
     let request_path = request.uri().path().to_string();
     let request_headers = request.headers().clone();
-    let custom_attrs = extract_custom_trace_attributes(
-        &request_headers,
-        tracing_config
-            .as_ref()
-            .as_ref()
-            .and_then(|tracing| tracing.span_attribute_header_prefixes.as_deref()),
-    );
+    let mut header_prefixes: Option<&[String]> = None;
```
salmanap: I think we can merge this into a single utility that both llm.rs and agent_chat_completions.rs work off of. Else its duplicate code and will be maintained in two places.

crates/brightstaff/src/tracing/custom_attributes.rs
Spherrrical: I've added a comment inside the `tracing.rst` files under Notes & Examples:

Snippet:
- **Prefix must match exactly**: ``katanemo-`` does not match ``x-katanemo-`` headers.
- **Trailing dash is recommended**: Without it, ``x-katanemo`` would also match ``x-katanemo-foo`` and
  ``x-katanemofoo``.
- **Keys are always strings**: Values are captured as string attributes.

The documentation above shows the flow of prefix -> parsing -> attribute shown.

docs/source/guides/observability/tracing.rst
```diff
@@ -142,6 +142,109 @@ In your observability platform (Jaeger, Grafana Tempo, Datadog, etc.), filter tr
...
+
+You configure one or more header prefixes. Any incoming HTTP header whose name starts with one of these
+prefixes is captured as a span attribute. You can also provide static attributes that are always injected.
+
+- The **prefix is only for matching**, not the resulting attribute key.
+- The attribute key is the header name **with the prefix removed**, then hyphens converted to dots.
+
+.. note::
+
+   Custom span attributes are attached to LLM spans when handling ``/v1/...`` requests via ``llm_chat``. For orchestrator requests to ``/agents/...``,
```
salmanap, line 170: via llm_chat? I think thats an internal detail. Please fix this section,

salmanap (approved):

PR #721, opened 2026-02-02
use standard tracing and logging in brightstaff

adilhafeez:
### What changed

- **Replaced custom `TraceCollector` with `tracing-opentelemetry`** — deleted the custom HTTP-based span collection system and switched to the standard OpenTelemetry tracing
- **Switched tracig to use grpc** — `OTEL_TRACING_HTTP_ENDPOINT` → `OTEL_TRACING_GRPC_ENDPOINT`; removed the envoy `otel_collector_proxy` listener - not needed
- **New `ServiceNameOverrideExporter`** — supports per-span `service.name` overrides so Jaeger shows logical services (`plano(orchestrator)`, `plano(llm)`, etc.)
- **Span instrumentation** — replaced manual `SpanBuilder` + `trace_collector` parameter threading with `info_span!().instrument()` pattern; `request_id` auto-propagates via span context
- **Tracing config from `config.yaml`** — new `opentracing_grpc_endpoint` field with env var expansion and validation
- **Envoy update** Envoy 1.36→1.37 - needed for operation name update from envoy config
- **Envoy template** — OTEL cluster endpoint now templated from config
- **Demos** — replaced Open WebUI with AnythingLLM

### Sample log output

```
[2026-02-06 00:02:26.006][info] request_id=1b7f5f61-b03d3a28d791 handling request listener=travel_book...
[2026-02-06 00:02:26.286][info] request_id=1b7f5f61-b03d3a28d791 arch-orchestrator determined routes ...
[2026-02-06 00:02:26.286][info] request_id=1b7f5f61-b03d3a28d791 selected agents for execution count=2...
```

Traces now visible in Jaeger per-request:

~Also there is some degradation in naming and tags - I will spend some more time and fix them~ this is now fixed

image

Another update is that now we are also pushing logs to tracing so you can see logs for one request for example,

image

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -3,52 +3,94 @@ name: e2e tests
 on:
   push:
     branches:
-      - main  # Run tests on pushes to the main branch
+      - main
   pull_request:
 
-jobs:
-  e2e_tests:
-    runs-on: ubuntu-latest
+permissions:
+  contents: read
 
+# Shared env vars for all jobs that run tests
+env:
+  PLANO_DOCKER_IMAGE: katanemo/plano:e2e
+
+jobs:
+  # ──────────────────────────────────────────────
+  # Job 1: Build the Docker image once, with cache
+  # ──────────────────────────────────────────────
+  build:
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
-
-      # --- Disk inspection & cleanup section (added to free space on GitHub runner) ---
-      - name: Check disk usage before cleanup
-        run: |
-          echo "=== Disk usage before cleanup ==="
-          df -h
-          echo "=== Repo size ==="
-          du -sh .
+        uses: actions/checkout@v4
 
       - name: Free disk space on runner
         run: |
-          echo "=== Cleaning preinstalled SDKs and toolchains to free space ==="
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /usr/local/lib/android
-          sudo rm -rf /opt/ghc
-          # If you still hit disk issues, uncomment this to free more space.
-          # It just removes cached tool versions; setup-python will re-download what it needs.
-          # sudo rm -rf /opt/hostedtoolcache || true
-
-          echo "=== Docker cleanup (before our builds/compose) ==="
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
           docker system prune -af || true
           docker volume prune -f || true
 
-          echo "=== Disk usage after cleanup ==="
-          df -h
-      # --- End disk cleanup section ---
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build plano image (with GHA cache)
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: Dockerfile
+          load: true
+          tags: ${{ env.PLANO_DOCKER_IMAGE }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Save image as artifact
+        run: docker save ${{ env.PLANO_DOCKER_IMAGE }} -o /tmp/plano-image.tar
+
+      - name: Upload image artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: plano-image
+          path: /tmp/plano-image.tar
+          retention-days: 1
+
+  # ──────────────────────────────────────────────
+  # Job 2a: prompt_gateway tests
+  # ──────────────────────────────────────────────
+  test-prompt-gateway:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Free disk space on runner
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
+          docker system prune -af || true
+          docker volume prune -f || true
+
+      - name: Download plano image
+        uses: actions/download-artifact@v4
+        with:
+          name: plano-image
+          path: /tmp
+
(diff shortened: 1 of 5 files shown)


crates/brightstaff/src/handlers/llm.rs
```diff
@@ -96,13 +125,10 @@ pub async fn llm_chat(
         Ok(request) => request,
         Err(err) => {
             warn!(
-                "[PLANO_REQ_ID:{}] | FAILURE | Failed to parse request as ProviderRequestType: {}",
```
salmanap, line 99: We are losing the PLANO_XX_XX style attention headers. I think want structured way to separate critical parts of the request. Else all the logs look like one big blog and really hard to follow. We moved away from the lowercase version to this scheme to make it easier to follow. This impacts developer experience.

crates/brightstaff/src/handlers/pipeline_processor.rs
```diff
@@ -81,125 +91,21 @@ impl PipelineProcessor {
...
-            span_builder = span_builder.with_parent_span_id(parent_span_id);
-        }
-
-        let span = span_builder.build();
-        // MCP spans also use plano(filter) service name as they are part of filter operations
-        collector.record_span(operation_component::AGENT_FILTER, span);
-    }
-
     /// Process the filter chain of agents (all except the terminal agent)
+    #[instrument(
```
salmanap: I like this view - but how does this get batched and processed async?

crates/brightstaff/src/handlers/router_chat.rs
salmanap, line 1: The router call is a separate http call - believe it needs its own span for latency and errors. I see you removed it here.

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -87,186 +92,99 @@ impl StreamProcessor for ObservableStreamProcessor {
...
-                },
-            });
-
-            // Add time to first token as a span event
-            // Calculate the timestamp by adding ttft duration to span start time
-            if let Ok(start_time_nanos) = self.span.start_time_unix_nano.parse::<u128>() {
-                // Convert ttft from milliseconds to nanoseconds and add to start time
-                let event_timestamp = start_time_nanos + (ttft * 1_000_000);
-                let mut event =
-                    Event::new(llm::TIME_TO_FIRST_TOKEN_MS.to_string(), event_timestamp);
```
salmanap, line 128: We shouldn't remove the TTFT and llm-native metrics with this release. That would be a developer experience regression. No?

crates/brightstaff/src/tracing/service_name_exporter.rs
salmanap, line 1: This feels computationally expensive to recompute all the service names for each of the Spans. I presume these are in the async non-block path? It feels like this is the new TraceCollector?

salmanap (commented):
Left a couple of comments, and one meta comment is that I think we should remove all internal tracing attributes like thread_id, code.line_number unless debug is turned on. Else its all the internal implementation detail that the developer doesn't care about and will be confusing to him/her.

crates/brightstaff/src/handlers/router_chat.rs
adilhafeez: I was trying to figure this out - we don't log traces from 12001 and thats why we're not seeing router calls.  Will put this back.

crates/brightstaff/src/handlers/utils.rs
adilhafeez: That was not intended - will put it back

crates/brightstaff/src/tracing/service_name_exporter.rs
adilhafeez: This is happening in non-blocking path and its running a o(n) loop over number of spans collected look at line #64 where hashmap for all traces is constructed making it a o(1) call.

```
pub struct ServiceNameOverrideExporter {
    /// Map from service name → pre-created OTLP exporter (behind tokio Mutex
    /// because `SpanExporter::export` takes `&self` and the future must be Send).
    exporters: HashMap<String, Mutex<opentelemetry_otlp::SpanExporter>>,
}
```

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -34,6 +35,19 @@ fn full<T: Into<Bytes>>(chunk: T) -> BoxBody<Bytes, hyper::Error> {
         .boxed()
 }
 
+/// Adapter to inject OpenTelemetry trace context into Hyper HeaderMap
```
salmanap: didn't understand what this does.

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -345,12 +365,18 @@ async fn llm_chat_inner(
     // remove content-length header if it exists
     request_headers.remove(header::CONTENT_LENGTH);
 
+    // Inject current LLM span's trace context so upstream spans are children of plano(llm)
```
salmanap, line 356: I think this is related to the comment above - didn't follow the flow here.

crates/brightstaff/src/router/plano_orchestrator.rs
```diff
@@ -2,20 +2,31 @@ use std::{collections::HashMap, sync::Arc};
...
+use opentelemetry::{global, propagation::Injector};
 use thiserror::Error;
 use tracing::{debug, info, warn};
 
 use crate::router::orchestrator_model_v1::{self};
 
 use super::orchestrator_model::OrchestratorModel;
 
+/// Adapter to inject OpenTelemetry trace context into Hyper HeaderMap
+struct HeaderMapInjector<'a>(&'a mut header::HeaderMap);
```
salmanap: same header map injector - if this is really needed then we can't have that in two places correct?

salmanap (commented):
Just one more comment on this header map injector which is repeated twice. Didn't quite follow the need and also there is duplicated code for it. If anything it should be moved out more centrally if multiple crates are using it.

PR #744, opened 2026-02-12
Run plano natively by default

adilhafeez:
## Summary

- Native mode is now the default (`planoai up`), Docker is opt-in (`--docker`)
- Pre-compiled binaries (Envoy, WASM plugins, brightstaff) auto-downloaded on first run
- Local build artifacts used when available (e.g. after `planoai build`)
- Config templates bundled in wheel — `pip install planoai && planoai up config.yaml` works without repo clone
- CI workflow to publish binaries on release + native smoke test


salmanap:
Before I do a line-level code review: Is the build command mandatory? Meaning if we don't build native then the developer can't run native? This presumes that they must have our source code locally. 

If the idea is for us to be a stand-alone binary (much like Envoy can be) then shouldn't we support that as a first-class option. Plus I am not sure if I have seen this --native flag in other open source projects that support both a dockerized container image and a binary. Can you share examples on the developer experience that might inform me here?

PR #764, opened 2026-02-17
sync CLI templates with demo configs via manifest + CI flow

### LINKED ISSUE #738: Sync CLI templates with demo templates
There needs to be a more consistent way to make sure our demo templates, available in the `demos` folder, are properly propagated to our CLI  to insure up-to-date configurations.

Spherrrical:
## Summary

- Add a template sync manifest at `cli/planoai/templates/template_sync_map.yaml` to map each CLI template to its corresponding demo `config.yaml`.
- Add `cli/planoai/template_sync.py` with a write-only sync path via `write_mapped_demo_configs`, plus manifest validation for missing/unknown template IDs and missing files.
- Run template sync in CI with `uv run python -m planoai.template_sync` in `.github/workflows/ci.yml` so mapped demo configs are written on every run.
- 
Resolves #738

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -91,4 +91,4 @@ source .venv/bin/activate
 planoai build  # No need for 'uv run' when activated
 ```
 
-**Note:** For end-user installation instructions, see the [plano documentation](https://docs.planoai.dev).
+**Note:** For end-user installation instructions, see the [Plano documentation](https://docs.planoai.dev).


.github/workflows/ci.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Install plano tools
         run: uv sync --extra dev
 
+      - name: Sync CLI templates to demos
```
salmanap, line 49: We should run this specific CI command on merging to main - else it might be superflouous for every PR.

PR #769, opened 2026-02-18
Add mock-based E2E tests and gate live tests to main/nightly

adilhafeez:
## Summary
- **New mock-based E2E test suite** (33 tests) using `pytest_httpserver` to simulate LLM provider responses — runs on every PR with zero secrets required
- **Gate all live E2E jobs** to main pushes + daily nightly cron schedule (`0 6 * * *`), so PRs no longer depend on API keys
- **Scope secrets** per CI job to only the keys each job actually needs (removed unused AZURE_API_KEY, AWS_BEARER_TOKEN_BEDROCK, ARCH_API_KEY from jobs that don't need them; added missing GROQ_API_KEY to test-prompt-gateway)
- **Relax exact-match LLM assertions** in live e2e tests to structural checks (`is not None` + `len > 0`) since LLM output is non-deterministic

## New test coverage (mock-based)
| File | Tests | What it covers |
|------|-------|----------------|
| `test_model_alias_routing.py` | 13 | Alias resolution, cross-provider protocol transformation, tool calls, thinking mode, error handling |
| `test_responses_api.py` | 13 | Passthrough, upstream translation, tools, mixed content, multi-turn state management |
| `test_streaming.py` | 7 | Streaming for all API shapes (OpenAI chat, Anthropic messages, Responses API) + cross-provider |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,7 +305,11 @@ jobs:
       - name: Run model alias routing tests
         env:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
+          GROQ_API_KEY: ${{ secrets.GROQ_API_KEY }}
           ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          AZURE_API_KEY: ${{ secrets.AZURE_API_KEY }}
+          AWS_BEARER_TOKEN_BEDROCK: ${{ secrets.AWS_BEARER_TOKEN_BEDROCK }}
           GROK_API_KEY: ${{ secrets.GROK_API_KEY }}
         run: |
           cd tests/e2e && bash run_model_alias_tests.sh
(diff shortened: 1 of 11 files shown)


salmanap (commented):
Couple of thoughts

1/ If the code path that we are testing is specific to an upstream LLM change, then we should have a way to test out the integration test in the PR. This way we can know with confidence that changes don't break the user experience when merging to main. Else it would be cut another PR and fix main. Not ideal

2/ I am not sure what value the mock tests create? If the idea is that we are testing our transformation logic and these upstream mock LLMs help us with that then that makes sense. But most of these tests are integration tests trying to see if the client (OpenAI, Anthropic) can process the message that plano sends back. In that case, I am not sure if there is a lot of value in creating a mock suite. 

3/ If we go down this route (barring the discussions above). I think we don't want to duplicate each test type like streaming, model routing, or responses_api. We just need a way to update the tests to use a mock LLM. Also, I don't think we need put these tests in the old archgw folder.

salmanap (changes requested):

Spherrrical:
Closing for inactivity

PR #771, opened 2026-02-19
Support filter chains on model listeners

### LINKED ISSUE #729: Support Filter Chains on type: model listeners (LLM provider proxy)
Hi, thanks for the project.
Right now it looks like filter_chain can only be attached to agents under type: agent listeners. Is it possible (or planned) to support Filter Chains on type: model listeners as well?

Our use case is to run input guardrails (and optionally non-streaming output checks) as a transparent gateway for OpenAI-compatible requests (Chat Completions first, later Responses and Anthropic Messages), without forcing users to route through an “agent” layer.

If this isn’t supported, what’s the recommended pattern for “model proxy + guardrails” without introducing an extra passthrough agent hop?

## Maybe like this?

```yaml
version: v0.3.0
filters:
  - id: content_guard
    url: http://localhost:10500
    type: http
  - id: audit_logger
    url: http://localhost:10501
    type: mcp
model_providers:
  - model: openai/gpt-5.2
    access_key: $OPENAI_API_KEY
    filter_chain:
      - content_guard
      - audit_logger
listeners:
  - type: model
    name: llm_gateway
    port: 12000
```

adilhafeez:
## Summary
- Adds `filter_chain` support on model-type listeners, allowing input guardrails on direct LLM proxy requests without an agent layer
- Reuses the existing `PipelineProcessor.process_filter_chain()` subsystem from agent listeners
- Validates filter_chain IDs in Python config generator

Closes #729

### DIFF
--- a/crates/brightstaff/src/handlers/llm.rs
+++ b/crates/brightstaff/src/handlers/llm.rs
@@ -272,55 +272,50 @@ async fn llm_chat_inner(
             let mut pipeline_processor = PipelineProcessor::default();
             let messages = client_request.get_messages();
             match pipeline_processor
-                .process_filter_chain(
-                    &messages,
-                    &temp_filter_chain,
-                    &mfc.agents,
-                    &request_headers,
-                )
+                .process_filter_chain(&messages, &temp_filter_chain, &mfc.agents, &request_headers)
                 .await
-                {
-                    Ok(filtered_messages) => {
-                        client_request.set_messages(&filtered_messages);
-                        info!(
-                            original_count = messages.len(),
-                            filtered_count = filtered_messages.len(),
-                            "filter chain processed successfully"
-                        );
-                    }
-                    Err(super::pipeline_processor::PipelineError::ClientError {
-                        agent,
-                        status,
-                        body,
-                    }) => {
-                        warn!(
-                            agent = %agent,
-                            status = %status,
-                            body = %body,
-                            "client error from filter chain"
-                        );
-                        let error_json = serde_json::json!({
-                            "error": "FilterChainError",
-                            "agent": agent,
-                            "status": status,
-                            "agent_response": body
-                        });
-                        let mut error_response = Response::new(full(error_json.to_string()));
-                        *error_response.status_mut() =
-                            StatusCode::from_u16(status).unwrap_or(StatusCode::BAD_REQUEST);
-                        error_response.headers_mut().insert(
-                            hyper::header::CONTENT_TYPE,
-                            "application/json".parse().unwrap(),
-                        );
-                        return Ok(error_response);
-                    }
-                    Err(err) => {
-                        warn!(error = %err, "filter chain processing failed");
-                        let err_msg = format!("Filter chain processing failed: {}", err);
-                        let mut internal_error = Response::new(full(err_msg));
-                        *internal_error.status_mut() = StatusCode::INTERNAL_SERVER_ERROR;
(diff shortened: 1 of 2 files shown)


crates/brightstaff/src/handlers/llm.rs
```diff
@@ -98,6 +105,8 @@ async fn llm_chat_inner(
     request_id: String,
     request_path: String,
     mut request_headers: hyper::HeaderMap,
+    filter_chain: Arc<Option<Vec<String>>>,
```
salmanap: I think the naming here could be a bit better - I think the filter_chain_agent is the one that actually does the processing correct? Shouldn't the representation of that object be in the filter_chain list object. I guess at first blush the two objects didn't make sense.

demos/filter_chains/model_listener_filter/config.yaml
```diff
@@ -0,0 +1,21 @@
...
+model_providers:
+  - model: openai/gpt-4o-mini
+    access_key: $OPENAI_API_KEY
+    default: true
+
+listeners:
+  - type: model
+    name: llm_gateway
+    port: 12000
+    filter_chain:
```
salmanap, line 17: I shared this feedback in the other branch - but I think you benefit from calling it req_filter_chain especially as you introduce the resp_filter_chain in the other PR.

crates/brightstaff/src/main.rs
```diff
@@ -80,11 +81,46 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
     let llm_providers = Arc::new(RwLock::new(llm_providers));
     let combined_agents_filters_list = Arc::new(RwLock::new(Some(all_agents)));
+
+    // Resolve model listener filter chain and agents at startup
+    let model_listener_count = plano_config
+        .listeners
+        .iter()
+        .filter(|l| l.listener_type == ListenerType::Model)
+        .count();
+    assert!(
```
salmanap: Can't we assert that in the policy config check? Why introduce this in bright_staff and fail late?

crates/brightstaff/src/main.rs
```diff
@@ -80,11 +81,46 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
...
+        "only one model listener is allowed, found {}",
+        model_listener_count
+    );
+    let model_listener = plano_config
+        .listeners
+        .iter()
+        .find(|l| l.listener_type == ListenerType::Model);
+    let model_filter_chain: Arc<Option<Vec<String>>> =
+        Arc::new(model_listener.and_then(|l| l.filter_chain.clone()));
+    let model_filter_agents: Arc<HashMap<String, Agent>> = Arc::new(
```
salmanap: Can't we compress this as one struct? I was having a hard time understanding why we have two objects that we pass around everywhere?

crates/brightstaff/src/handlers/llm.rs
adilhafeez: thats taken care of next pr - wanted to keep it simple

crates/brightstaff/src/main.rs
adilhafeez: thats definitely good feedback - added check in config generator

demos/filter_chains/model_listener_filter/config.yaml
adilhafeez: not changing filter_chain name in this PR

Issue #706, opened 2026-01-27
Structured and Consistent Error Messages from Plano


salmanap:
Today, Plano errors aren't not consistently structured across the code base. The format and content of our messages is fragmented as called out here: https://github.com/katanemo/plano/pull/696#discussion_r2729168575. This issue will improve our error logging across all components.

salmanap:
I think we'd want to cover all 5xx and 4xx scenarios and make sure we create a node in our docs to talk about our errors. Btw, filter chain errors and routing errors should be where we house the "details" section with structured data

PR #772, opened 2026-02-19
[ISSUE 706]: Standardize returned errors from Plano

syedhashmi:
**Overview**
First PR related to Issue #706 . This PR sets up the base workflow for how standardized errors will be returned from Plano and does the cleanup in llm.rs. Next PRs will clean up rest of the paths.

**Testing done**
Using curl commands, verified that before the fix, returned errors were inconsistent (some even text). After the fix, returned error is standardized json as mentioned in the issue.

Here is one example of the curl command used to hit the error route:

`curl -X POST http://localhost:12000/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "abc/xyz",
    "messages": [
      {
        "role": "user",
        "content": "What is the weather like in San Francisco?"
      }
    ],
    "tools": [
      {
        "type": "functions",
        "function": {
          "name": "get_weather",
          "description": "Get the current weather in a given location",
          "parameters": {
            "type": "object",
            "properties": {
              "location": {
                "type": "string",
                "description": "The city and state, e.g. San Francisco, CA"
              },
              "unit": {
                "type": "string",
                "enum": ["celsius", "fahrenheit"],
                "description": "The temperature unit to use"
              }
            },
            "required": ["location"]
          }
        }
      }
    ],
    "tool_choice": "auto"
  }'`

Before the change , the returned error:
Model 'abc/xyz' not found in configured providers%

After the change:
`
{
  "error": {
    "code": "ModelNotFound",
    "message": "The requested model 'abc/xyz' does not exist",
    "details": {
      "rejected_model_id": "abc/xyz"
    }
  }
}
`

### DIFF
--- a/crates/brightstaff/src/handlers/agent_chat_completions.rs
+++ b/crates/brightstaff/src/handlers/agent_chat_completions.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use bytes::Bytes;
+use common::errors::BrightStaffError;
 use common::llm_providers::LlmProviders;
 use hermesllm::apis::OpenAIMessage;
 use hermesllm::clients::SupportedAPIsFromClient;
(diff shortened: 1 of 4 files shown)


syedhashmi:
@adilhafeez / @salmanap : Updated the Pr with:
- Standardized paths for agent chat completion
- Added tests for translation logic
- Cleanup errors as per previous comments

PR #799, opened 2026-03-05
fix: make upstream timeout configurable, default to 300s

### LINKED ISSUE #787: WASM filter hardcoded 30s timeout breaks agentic/tool-use workloads
## Summary

Plano's WASM filters hardcode 30-second timeouts on all outbound HTTP calls, making Plano unusable for agentic workloads where LLM calls with tool use regularly exceed 30 seconds — despite [Plano's docs explicitly supporting agent architectures](https://docs.planoai.dev/concepts/agents.html).

## Root Cause

Four constants in `crates/common/src/consts.rs`:

```rust
pub const ARCH_FC_REQUEST_TIMEOUT_MS: u64 = 30000;
pub const DEFAULT_TARGET_REQUEST_TIMEOUT_MS: u64 = 30000;
pub const API_REQUEST_TIMEOUT_MS: u64 = 30000;
pub const MODEL_SERVER_REQUEST_TIMEOUT_MS: u64 = 30000;
```

These are used in two ways:

1. **Request header** — `stream_context.rs` and `http_context.rs` set `x-envoy-upstream-rq-timeout-ms: 30000` on outbound calls, overriding Envoy's route-level `timeout: 300s`.

2. **dispatch_http_call** — the WASM `dispatch_http_call` function is called with a `Duration` based on these constants, hardcoding the timeout in the WASM binary itself.

The YAML listener `timeout` field only affects the inbound Envoy route timeout and has no effect on the WASM filter's outbound calls.

## Observed Behavior

When an LLM call takes >30s (common with tool use / function calling):

1. WASM filter dispatches HTTP call with 30s timeout
2. Envoy cancels the upstream connection after 30s
3. The upstream (brightstaff or LLM provider) may still be processing
4. LLM response arrives but the connection is already severed
5. Client receives `hyper::Error(IncompleteMessage)` or 504

The LLM call completes and tokens are consumed, but the response is discarded because the proxy chain is broken.

## Reproduce

```yaml
# plano_config.yaml
version: v0.3.0
model_providers:
  - model: openai/gpt-4o
    access_key: $OPENAI_API_KEY
listeners:
  - type: model
    name: llm-gateway
    port: 12000
    timeout: 300s   # does NOT help
```

```python
# test_timeout.py — pip install openai
import openai, time

client = openai.OpenAI(
    base_url="http://localhost:12000/v1",
    api_key="unused",
)

tools = [{
    "type": "function",
    "function": {
        "name": "slow_lookup",
        "description": "Look up data from a slow database",
        "parameters": {
            "type": "object",
            "properties": {"query": {"type": "string"}},
            "required": ["query"]
        }
    }
}]

# Step 1: model requests tool calls
response = client.chat.completions.create(
    model="openai/gpt-4o",
    messages=[{"role": "user", "content": "Look up the population of every Scandinavian country using slow_lookup for each one, then summarize."}],
    tools=tools,
)

# Step 2: simulate slow tool execution (agent loop)
messages = [{"role": "user", "content": "Look up the population of every Scandinavian country using slow_lookup for each one, then summarize."}]
messages.append(response.choices[0].message)

for tc in response.choices[0].message.tool_calls:
    time.sleep(8)  # 4-5 tools * 8s = 32-40s total
    messages.append({
        "role": "tool",
        "tool_call_id": tc.id,
        "content": f"Population: 5,400,000"
    })

# Step 3: send tool results back through Plano
try:
    final = client.chat.completions.create(
        model="openai/gpt-4o",
        messages=messages,
        tools=tools,
    )
    print("Success:", final.choices[0].message.content[:200])
except Exception as e:
    print(f"Failed: {e}")
```

In production, PydanticAI agents with 3-4 tool calls consistently fail at ~30s through Plano but work fine calling providers directly.

## Suggested Fix

Either:
1. Read the listener `timeout` config in the WASM filter and use it for outbound calls
2. Add a dedicated `upstream_timeout` config field
3. Remove the hardcoded header and let Envoy's route-level `timeout: 300s` govern

For agentic workloads, 120-300s is a reasonable default.

## Environment

- `katanemo/plano:0.4.8`
- Kubernetes
- Models: OpenAI GPT-4o, Anthropic Claude Sonnet via PydanticAI agents

syedhashmi:
Placeholder PR for Adil to review the timeout related changes.

fixes https://github.com/katanemo/plano/issues/787


cli/planoai/config_generator.py
```diff
@@ -466,6 +466,15 @@ def validate_and_render_schema():
         "upstream_tls_ca_path", "/etc/ssl/certs/ca-certificates.crt"
     )
 
+    upstream_timeout_ms = overrides.get("upstream_timeout_ms")
```
salmanap: not sure why we we have an upstream_timeout_rs field, when the model_listener object already has a timeout field. Can you elaborate a bit more?

config/plano_config_schema.yaml
```diff
@@ -265,6 +265,10 @@ properties:
         type: boolean
       use_agent_orchestrator:
         type: boolean
+      upstream_timeout_ms:
```
salmanap: Same as above. I don't think we need this field, especially if we already support a timeout field for model_listener objects. Please review more carefully

crates/prompt_gateway/src/http_context.rs
salmanap, line 1: As mentioned over the zoom call - we don't need any changes to the prompt_gateway side of things. The issue talked about how the llm_gateway was the one timing out and the developer may have had a tool call scenario that could have taken longer.

salmanap (commented):
Some comments on the PR. Also, I think we would need a test case here - else its hard to tell if the change in timeout is actually working. Lastly, I don't follow from the issues request how the user simulated a timeout. From the looks of that code, I see that he is adding a time.sleep on local code, which has no implication on the response from an upstream LLM

adilhafeez:
So there are at multiple timeouts we are talking here,

- Connection timeout — Envoy cluster connect_timeout (5s default)
- Route timeout — Envoy route timeout (inbound client request lifecycle)
- WASM outbound call timeout — dispatch_http_call Duration + x-envoy-upstream-rq-timeout-ms
- Brightstaff outbound call timeout — reqwest client timeout (currently none)

For default values we should use sensible defaults for connection and request timeouts. And a developer should be able to modify them using overrides section in config.yaml. And defaults should be defined centrally somewhere and let's discuss their values here.

For example here is what I think default should be,

- connection_timeout: 1s
- request_timeout: 120s

salmanap:
@syedhashmi when are we wrapping this up? We need to get this over the finish line please.

cli/planoai/config_generator.py
syedhashmi: Updated to use existing per listener timeout.

config/plano_config_schema.yaml
syedhashmi: Updated.

Spherrrical:
Closing for inactivity

PR #808, opened 2026-03-06
Support for Codex via Plano

Spherrrical:
## Summary
- Add end-to-end Codex CLI agent support in the Plano CLI, including native-mode compatibility checks so `planoai cli-agent codex ...` works when Plano is running natively.
- Extend provider compatibility and request normalization so xAI works with Responses API flows, including native `/v1/responses` routing and tool-shape normalization to avoid `custom` tool rejection.
- Improve OpenAI/Responses request+stream transforms and add a runnable Codex routing demo with docs/config/scripts for local validation.

## Test plan
- [x] `cargo test -p hermesllm test_xai_uses_responses_api_for_responses_clients`
- [x] `cargo test -p hermesllm test_responses_api_targets_xai_native_responses_endpoint`
- [x] `cargo test -p brightstaff test_normalize_xai_custom_tool_to_function`
- [x] `cargo test -p brightstaff test_normalize_xai_non_custom_tool_unchanged`
- [x] `cargo test -p hermesllm test_request_deserializes_web_search_tool_alias`
- [x] `cargo test -p llm_gateway --lib`
- [x] `uv run pytest -q test/test_cli_agent.py`
- [x] Manual smoke test: `planoai --version` resolves to local editable install and reports expected version.
- [x] Manual smoke test: native Plano up + Codex routing path validated via gateway logs and xAI 422 regression fixed in hot path.

### DIFF
--- a/crates/hermesllm/src/apis/openai_responses.rs
+++ b/crates/hermesllm/src/apis/openai_responses.rs
@@ -1036,6 +1036,30 @@ pub struct ListInputItemsResponse {
 // ProviderRequest Implementation
 // ============================================================================
 
+fn append_input_content_text(buffer: &mut String, content: &InputContent) {
+    match content {
+        InputContent::InputText { text } => buffer.push_str(text),
+        InputContent::InputImage { .. } => buffer.push_str("[Image]"),
+        InputContent::InputFile { .. } => buffer.push_str("[File]"),
+        InputContent::InputAudio { .. } => buffer.push_str("[Audio]"),
+    }
+}
+
+fn append_content_items_text(buffer: &mut String, content_items: &[InputContent]) {
+    for content in content_items {
+        // Preserve existing behavior: each content item is prefixed with a space.
+        buffer.push(' ');
+        append_input_content_text(buffer, content);
+    }
+}
+
+fn append_message_content_text(buffer: &mut String, content: &MessageContent) {
+    match content {
+        MessageContent::Text(text) => buffer.push_str(text),
+        MessageContent::Items(content_items) => append_content_items_text(buffer, content_items),
+    }
+}
+
 impl ProviderRequest for ResponsesAPIRequest {
     fn model(&self) -> &str {
         &self.model
@@ -1057,29 +1081,7 @@ impl ProviderRequest for ResponsesAPIRequest {
                 match item {
                     InputItem::Message(msg) => {
                         let mut extracted = String::new();
-                        match &msg.content {
-                            MessageContent::Text(text) => extracted.push_str(text),
-                            MessageContent::Items(content_items) => {
-                                for content in content_items {
-                                    // Preserve existing behavior: each content item is prefixed with a space.
-                                    extracted.push(' ');
-                                    match content {
-                                        InputContent::InputText { text } => {
-                                            extracted.push_str(text)
-                                        }
-                                        InputContent::InputImage { .. } => {
-                                            extracted.push_str("[Image]")
-                                        }
-                                        InputContent::InputFile { .. } => {
-                                            extracted.push_str("[File]")
-                                        }
-                                        InputContent::InputAudio { .. } => {
-                                            extracted.push_str("[Audio]")
-                                        }
-                                    }
-                                }
-                            }
-                        }
+                        append_message_content_text(&mut extracted, &msg.content);
                         extracted
                     }
                     _ => String::new(),
@@ -1091,29 +1093,7 @@ impl ProviderRequest for ResponsesAPIRequest {
                     if let InputItem::Message(msg) = item {
                         // Preserve existing behavior: each message is prefixed with a space.
                         extracted.push(' ');
-                        match &msg.content {
-                            MessageContent::Text(text) => extracted.push_str(text),
-                            MessageContent::Items(content_items) => {
-                                for content in content_items {
-                                    // Preserve existing behavior: each content item is prefixed with a space.
-                                    extracted.push(' ');
-                                    match content {
-                                        InputContent::InputText { text } => {
-                                            extracted.push_str(text)
-                                        }
-                                        InputContent::InputImage { .. } => {
-                                            extracted.push_str("[Image]")
-                                        }
-                                        InputContent::InputFile { .. } => {
-                                            extracted.push_str("[File]")
-                                        }
-                                        InputContent::InputAudio { .. } => {
-                                            extracted.push_str("[Audio]")
-                                        }
-                                    }
-                                }
-                            }
-                        }
+                        append_message_content_text(&mut extracted, &msg.content);
                     }
                 }
                 extracted
--- a/crates/hermesllm/src/clients/endpoints.rs
+++ b/crates/hermesllm/src/clients/endpoints.rs
@@ -185,7 +185,7 @@ impl SupportedAPIsFromClient {
                 // For Responses API, check if provider supports it, otherwise translate to chat/completions
                 match provider_id {
                     // Providers that support /v1/responses natively
-                    ProviderId::OpenAI => route_by_provider("/responses"),
+                    ProviderId::OpenAI | ProviderId::XAI => route_by_provider("/responses"),
                     // All other providers: translate to /chat/completions
                     _ => route_by_provider("/chat/completions"),
                 }
@@ -656,7 +656,7 @@ mod tests {
     }
 
     #[test]
-    fn test_responses_api_targets_xai_chat_completions_endpoint() {
+    fn test_responses_api_targets_xai_native_responses_endpoint() {
         let api = SupportedAPIsFromClient::OpenAIResponsesAPI(OpenAIApi::Responses);
         assert_eq!(
             api.target_endpoint_for_provider(
@@ -666,7 +666,7 @@ mod tests {
                 false,
                 None
             ),
-            "/v1/chat/completions"
+            "/v1/responses"
         );
     }
 }
--- a/crates/hermesllm/src/providers/id.rs
+++ b/crates/hermesllm/src/providers/id.rs
@@ -166,10 +166,11 @@ impl ProviderId {
                 SupportedAPIsFromClient::OpenAIChatCompletions(_),
             ) => SupportedUpstreamAPIs::OpenAIChatCompletions(OpenAIApi::ChatCompletions),
 
-            // OpenAI Responses API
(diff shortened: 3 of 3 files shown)


cli/planoai/core.py
```diff
@@ -147,26 +146,48 @@ def stop_docker_container(service=PLANO_DOCKER_NAME):
...
+    if isinstance(listeners, dict):
+        egress_config = listeners.get("egress_traffic", {})
+        host = egress_config.get("host") or egress_config.get("address") or "127.0.0.1"
+        port = egress_config.get("port", 12000)
+        return host, port
+
+    if isinstance(listeners, list):
+        for listener in listeners:
+            if listener.get("type") in ["model", "model_listener"]:
+                host = listener.get("host") or listener.get("address") or "127.0.0.1"
```
salmanap: I think we should settle on 127.0.0.1 or 0.0.0.0. I think we have an inconsistent default in this case for the proxy interface.

cli/test/test_cli_agent.py
salmanap, line 1: I don't think you need this file - unless its part of our CI/CD pipeline. In this instance, manually testing the CLI is okay.

crates/brightstaff/src/handlers/llm.rs
```diff
@@ -551,3 +557,89 @@ async fn get_provider_info(
         (hermesllm::ProviderId::OpenAI, None)
     }
 }
+
+fn normalize_xai_responses_tool(tool: ResponsesTool, idx: usize) -> ResponsesTool {
```
salmanap: Two things 

1. why is this being done just for xAI when xAI supports a v1/chat/completions compatible API and hermes offers v1/responses to v1/chat/completions transformations

2. why is this happening in brighstaff? All transformation logic should go into hermes where we check for ProviderId type checks. Brighstaff is strictly for controlling traffic not transforming it

crates/brightstaff/src/state/mod.rs
```diff
@@ -130,6 +130,7 @@ pub fn extract_input_items(input: &InputParam) -> Vec<InputItem> {
                 }]),
             })]
         }
+        InputParam::SingleItem(item) => vec![item.clone()],
```
salmanap, line 133: This is touching our v1/responses implementation. Do we need to write a test for this? Are there any backwards compatibility issues that this would create?

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
```diff
@@ -255,8 +257,38 @@ impl ResponsesAPIStreamBuffer {
...
     pub fn finalize(&mut self) {
+        // Idempotent finalize: avoid duplicate response.completed loops.
+        if self.finalized {
+            return;
+        }
+        self.finalized = true;
+
         let mut events = Vec::new();
 
+        // Ensure lifecycle prelude is emitted even if finalize is triggered
```
salmanap, line 268: I am not sure I follow this code path. What specific thing was breaking if this change wasn't introduced. Can you share the payload this would fix. Feels like a hack and I am wondering if there is a structural problem with our resposne_api_streaming_buffer implementation

crates/hermesllm/src/apis/openai_responses.rs
```diff
@@ -166,26 +178,32 @@ pub enum MessageRole {
...
     },
     /// File input via URL
+    #[serde(rename = "input_file", alias = "file")]
     InputFile { file_url: String },
     /// Audio input
     InputAudio {
         data: Option<String>,
         format: Option<String>,
     },
+    /// Forward-compat fallback for unknown content parts.
```
salmanap: We should have a very specific reason for Unknown - this would mask future transformation issues. Can you just remove this line of code and test and see what really breaks?

crates/hermesllm/src/apis/openai_responses.rs
```diff
@@ -1031,6 +1057,27 @@ impl ProviderRequest for ResponsesAPIRequest {
     fn extract_messages_text(&self) -> String {
         match &self.input {
             InputParam::Text(text) => text.clone(),
+            InputParam::SingleItem(item) => {
```
salmanap, line 1079: This is an interesting block of code. This repeats the functionality that extract_messages_text already has for InputParam::Items. We should make sure we create a helper method such that a SingleItem and Items use the exact same content_items.iter().fold() - else these two code paths will diverge and we will introduce subtle bugs in our transformation logic

crates/llm_gateway/src/stream_context.rs
```diff
@@ -1225,6 +1231,51 @@ impl HttpContext for StreamContext {
...
+    }
+    if !matches!(resolved_api, SupportedUpstreamAPIs::OpenAIResponsesAPI(_)) {
+        return;
+    }
+    if let ProviderRequestType::ResponsesAPIRequest(responses_req) = request {
+        normalize_responses_tools_for_xai(responses_req);
+    }
+}
+
+fn normalize_responses_tools_for_xai(req: &mut ResponsesAPIRequest) {
```
salmanap: This is the second place we are normalizing for xAI and is a bad smell. Please see comment above that if we are introducing a v1/responses API for xAI then all transformation logic should be handled in hermes

crates/hermesllm/src/transforms/response_streaming/to_openai_streaming.rs
```diff
@@ -512,19 +512,12 @@ impl TryFrom<ChatCompletionsStreamResponse> for ResponsesAPIStreamEvent {
...
-                // The buffer will handle the finish_reason and generate response.completed
-                return Ok(ResponsesAPIStreamEvent::ResponseOutputTextDelta {
-                    item_id: "".to_string(), // Buffer will fill this
-                    output_index: choice.index as i32,
-                    content_index: 0,
-                    delta: "".to_string(), // Empty delta signals completion
-                    logprobs: vec![],
-                    obfuscation: None,
-                    sequence_number: 0, // Buffer will fill this
+                return Ok(ResponsesAPIStreamEvent::Done {
```
salmanap, line 519: This seems very strange. In my prior implementation we should to build the final buffers based on what we accumulated. This feels like a tricky change. Can you please share the bug that our prior implementation caused for this change?

crates/brightstaff/src/state/mod.rs
Spherrrical: the utility itself remains compatible (same normalization as before). The main change in this PR is xAI Responses traffic will now routing through the existing Responses -> ChatCompletions path (suggestion from you above). I also added provider-specific normalization in hermes to avoid sending deprecated live-search payloads to xAI chat upstream, because that was the main reason for refactoring into brightstaff instead of hermes.

crates/hermesllm/src/apis/streaming_shapes/responses_api_streaming_buffer.rs
Spherrrical: When upstream is chat.completions and client is v1/responses, we convert streaming chunks into Responses events. Before finalization depended on seeing wire `[DONE]` in add_transformed_event().

But some upstream streams can end with a finish_reason chunk and never deliver [DONE] (or [DONE] arrives late/gets dropped). In that case we emitted deltas, but never emitted terminal lifecycle events (*.done + response.completed).

Heres a sample payload that would affect this:
```
data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":1234567890,"model":"gpt-4o","choices":[{"index":0,"delta":{"role":"assistant","content":"Hello"},"finish_reason":null}]}

data: {"id":"chatcmpl-123","object":"chat.completion.chunk","created":1234567890,"model":"gpt-4o","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}
```

Without this patch, that sequence can end with:
- response.created
- response.in_progress
- response.output_text.delta and not:
- response.output_text.done
- response.output_item.done
- response.completed

so downstream clients can hang waiting for completion or see incomplete response lifecycle

crates/llm_gateway/src/stream_context.rs
Spherrrical: Resolved.

cli/planoai/core.py
Spherrrical: Resolved to 0.0.0.0

cli/test/test_cli_agent.py
Spherrrical: Good catch, was using this file as a local testing tool for a coding agent

crates/brightstaff/src/handlers/llm.rs
Spherrrical: 1. Resolved to xAI to use v1/chat/completions (previously it broke, but I found a solution)
2. Moved the implementation to hermes instead of brightstaff.

crates/hermesllm/src/apis/openai_responses.rs
Spherrrical: There were issues previously, but removing it now does not break anything - resolved this.

crates/hermesllm/src/transforms/response_streaming/to_openai_streaming.rs
Spherrrical: As one of the changes in the streaming shape, payloads like:

```
data: {"id":"chatcmpl-123","choices":[{"index":0,"delta":{"content":"Hello"},"finish_reason":null}]}

data: {"id":"chatcmpl-123","choices":[{"index":0,"delta":{},"finish_reason":"stop"}]}
```

would not emit the correct *.completed/*.output_item.done tags, which would contionously loop a response (trace logs show that it would repeatedly send over and over again). This change adds idempotency: so if [DONE] also arrives, finalization still happens only once.

crates/hermesllm/src/apis/openai_responses.rs
Spherrrical: Resolved to a helper function.

crates/hermesllm/src/apis/openai_responses.rs
```diff
@@ -1029,45 +1050,59 @@ impl ProviderRequest for ResponsesAPIRequest {
     }
 
     fn extract_messages_text(&self) -> String {
+        fn content_items_to_text(content_items: &[InputContent]) -> String {
```
salmanap: Why do you have an inline function within another function?

Spherrrical: Resolved.

salmanap (approved):
There are still some nested functions that create for some tech debt. But I will be in that code path and i'll fix those separately.

PR #809, opened 2026-03-09
Run demos without Docker

adilhafeez:
## Summary
- Custom Python agents in demos now run natively via `uv run` instead of Docker containers
- Docker is kept as an optional dependency (`--with-ui` flag) for third-party UIs (AnythingLLM, Open WebUI) and tracing (Jaeger)
- Created `start_agents.sh` scripts for 4 demos (weather_forecast, travel_agents, multi_agent_crewai_langchain, multi_turn_rag), fixed the existing one in http_filter
- Updated `config.yaml` files to use `localhost` instead of Docker container hostnames (multi_agent_crewai_langchain, http_filter)
- Updated all 12 demo `run_demo.sh` scripts and 7 READMEs to reflect the new workflow

## Usage
```bash
# Run demo without Docker (default)
./run_demo.sh

# Run demo with optional UI services
./run_demo.sh --with-ui

# Stop demo
./run_demo.sh down
```

### DIFF
--- a/demos/llm_routing/preference_based_routing/README.md
+++ b/demos/llm_routing/preference_based_routing/README.md
@@ -20,14 +20,14 @@ Then open AnythingLLM at http://localhost:3001/
 
 Or start manually:
 
-1. Start Plano
-```bash
-planoai up config.yaml
-```
-
-2. (Optional) Start AnythingLLM and Jaeger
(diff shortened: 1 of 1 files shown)


salmanap (approved):
These changes look really good - have you spot checked a few demos?

PR #820, opened 2026-03-11
Unified overrides for custom router and orchestrator models

adilhafeez:
- Replace separate `routing` and `orchestration` config sections with `overrides.router_model` and `overrides.orchestrator_model`
- Add `plano` provider interface (alias for `arch`) for self-hosted models
- Strip provider prefix when resolving model names for upstream requests
- Update docs, demos, and schema

### DIFF
--- a/docs/source/resources/includes/plano_config_full_reference_rendered.yaml
+++ b/docs/source/resources/includes/plano_config_full_reference_rendered.yaml
@@ -110,7 +110,7 @@ model_providers:
   provider_interface: plano
 - internal: true
   model: Plano-Orchestrator
-  name: plano-orchestrator
+  name: plano/orchestrator
   provider_interface: plano
 prompt_targets:
 - description: Get current weather at a location.


cli/planoai/config_generator.py
```diff
@@ -403,12 +399,21 @@ def validate_and_render_schema():
...
+    orchestrator_model_id = (
+        orchestrator_model.split("/", 1)[1]
+        if "/" in orchestrator_model
+        else orchestrator_model
+    )
+    if orchestrator_model_id not in model_name_set:
         updated_model_providers.append(
             {
                 "name": "plano-orchestrator",
                 "provider_interface": "arch",
```
salmanap: make this plano/plano-orchestrator

crates/common/src/configuration.rs
```diff
@@ -265,7 +263,15 @@ impl LlmProviderType {
     /// Get the ProviderId for this LlmProviderType
     /// Used with the new function-based hermesllm API
     pub fn to_provider_id(&self) -> hermesllm::ProviderId {
-        hermesllm::ProviderId::try_from(self.to_string().as_str())
+        // Plano provider uses the same interface as Arch
+        let provider_str = match self {
+            LlmProviderType::Plano => "arch",
```
salmanap: Just make that plano and create a cluster for it in the envoy config file?

docs/source/guides/orchestration.rst
```diff
@@ -335,6 +335,89 @@ Combine RAG agents for documentation lookup with specialized troubleshooting age
...
+
+   Use the model name matching your ``--served-model-name``:
+
+   .. code-block:: yaml
+
+       overrides:
+         orchestrator_model: plano/katanemo/Plano-Orchestrator-4B
+
+       model_providers:
+         - model: plano/katanemo/Plano-Orchestrator-4B
```
salmanap: just plano/Plano-Orchestrator is fine.

cli/planoai/config_generator.py
```diff
@@ -403,12 +399,21 @@ def validate_and_render_schema():
...
-                "model": "Plano-Orchestrator",
+                "model": orchestrator_model_id,
```
salmanap, line 415: we should standardize on the model field. I am not sure how the name is different from the model field.

config/plano_config_schema.yaml
```diff
@@ -271,6 +273,12 @@ properties:
       upstream_tls_ca_path:
         type: string
         description: "Path to the trusted CA bundle for upstream TLS verification. Default is '/etc/ssl/certs/ca-certificates.crt'."
+      router_model:
```
salmanap: how about llm_routing_model and agent_orchestration_model

docs/source/guides/orchestration.rst
adilhafeez: thats actually name of the model from hugging face

cli/planoai/config_generator.py
adilhafeez: updated

config/plano_config_schema.yaml
adilhafeez: its more descriptive and better - updating

PR #822, opened 2026-03-12
add output filter chain

adilhafeez:


crates/brightstaff/src/handlers/llm.rs
```diff
@@ -47,6 +48,8 @@ pub async fn llm_chat(
     state_storage: Option<Arc<dyn StateStorage>>,
     filter_chain: Arc<Option<Vec<String>>>,
     filter_agents: Arc<HashMap<String, Agent>>,
+    output_filter_chain: Arc<Option<Vec<String>>>,
+    output_filter_agents: Arc<HashMap<String, Agent>>,
```
salmanap: what's the difference between output_filter_agents and output_filter_chains? Is one the definition and the other is the execution. If so,  perhaps the names could be a bit more expressive?

demos/filter_chains/pii_anonymizer/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+model_providers:
+  - model: openai/gpt-4o-mini
+    access_key: $OPENAI_API_KEY
+    default: true
+
+listeners:
+  - type: model
+    name: llm_gateway
+    port: 12000
+    filter_chain:
```
salmanap: I think we should just call it req_filter_chain and resp_filter_chain. Just more declerative and easier to understand because the traffic is bi-directional.

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -277,6 +281,286 @@ where
     }
 }
 
+/// Extract content text from an SSE chunk line (the JSON part after "data: ").
+/// Returns the content string and whether it was found.
+fn extract_sse_content(json_str: &str) -> Option<String> {
```
salmanap: This won't work for non v1/chat/completions payloads. the SSE events are bound by the API shape. Do we need these changes in this PR?

demos/filter_chains/pii_anonymizer/config.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - pii_anonymizer
+    output_filter_chain:
+      - pii_deanonymizer
```
salmanap, line 25: where is the example for this?

crates/brightstaff/src/handlers/utils.rs
```diff
@@ -277,6 +281,286 @@ where
...
+        .insert(
+            "content".to_string(),
+            serde_json::Value::String(new_content.to_string()),
+        );
+    serde_json::to_string(&value).ok()
+}
+
+/// Process an SSE chunk through the output filter chain.
+/// Parses each `data: ` line, extracts content, sends through filters, and reconstructs.
+async fn filter_sse_chunk(
```
salmanap: This has some trickiness to it - meaning what's the contract we expect to send to the filter chain? In the case of our agent filter chains its essentially the full payload as we get from the client and they must implement a v1/chat/completions. 

But in this case, it looks like we are creating our own shape to send to the filter chain for LLM and re-constructing it to send to upstream?

How should we reconcile these things? The concepts look similar but they feel different. Now I realize that the output is more tricky given the streaming response. But wouldn't be helpful to apply some consistency in how we approach this?

crates/brightstaff/src/handlers/llm.rs
adilhafeez: I've renamed to input_filters and output_filters

demos/filter_chains/pii_anonymizer/config.yaml
adilhafeez: renamed to input_filters and output_filters

demos/filter_chains/pii_anonymizer/config.yaml
adilhafeez: :) - here in this folder look at README.md file

crates/brightstaff/src/handlers/utils.rs
adilhafeez: I've updated code only support output filters for v1/chat/completions as supporting it for all api paths would require bigger change

salmanap: But my higher level question till remains - why are we extract SSE events and not forwarding the event as is to the filter chain and let it mutate and send us back the same shape? Why should we get into this business of strictly supporting text from sse, when we can also support streaming HTTP and just act as a coordinator/orchestrator

PR #827, opened 2026-03-14
Model affinity for consistent model selection in agentic loops

### LINKED ISSUE #813: Support trajectory pinning for consistent model selection in agentic loops
## Summary

In agentic loops, the same conversation hits Plano's routing endpoint multiple times. Today each call re-evaluates routing independently, so the selected model can change mid-conversation. Session pinning ensures that once a model is selected for a session, subsequent routing calls in that session return the same model.

## How it works

The caller sends an `X-Session-Id` header in the routing request. Plano caches the routing decision keyed by that ID:

- **First call** with `X-Session-Id: abc` → run routing, select model, cache `abc → gpt-4o`
- **Subsequent calls** with `X-Session-Id: abc` → skip routing, return cached `gpt-4o`
- **Cache entries expire** via configurable TTL (default 30 min)
- **No `X-Session-Id` header** → routing runs fresh every time (current behavior, no breaking change)

## Request

```
POST /routing/v1/chat/completions
X-Session-Id: session-abc-123
Content-Type: application/json

{
  "messages": [...]
}
```

## Response

```json
{
  "model": "gpt-4o",
  "route": "quick",
  "session_id": "session-abc-123",
  "pinned": true,
  "trace_id": "abc123..."
}
```

`pinned: true` indicates the result came from cache. `pinned: false` on first routing decision.

## Implementation

- Extract `X-Session-Id` from request headers in the routing handler
- Add an in-memory TTL cache in `RouterService` keyed by session ID (e.g. `HashMap<String, (String, Instant)>` behind a mutex)
- Before calling `determine_route()`, check cache for a valid (non-expired) entry
- On cache miss, run routing and store the result
- TTL configurable via `routing.session_ttl_seconds` in plano config (default 1800)
- No database or external state needed

adilhafeez:
## Summary

Closes #813

Adds model affinity — a mechanism to pin the routing decision for the duration of an agentic session. When a client sends the `X-Model-Affinity` header, the first request routes normally and caches the result. All subsequent requests with the same affinity ID skip routing and reuse the cached model.

**Why it matters:** In agentic loops, successive prompts have different intents (tool selection → reasoning → code generation), so the router may switch models mid-session. This causes behavioral inconsistency, invalidates provider-side KV caches, and loses cached input token discounts.

### Changes

- **Header**: `X-Model-Affinity` (opt-in, no breaking change to existing clients)
- **Cache**: In-memory HashMap with TTL + oldest-entry eviction, background cleanup every 5 min
- **Config**: `routing.session_ttl_seconds` (default 600), `routing.session_max_entries` (default/max 10,000)
- **Both paths**: Works on direct LLM handler (`/v1/chat/completions`) and routing decision API (`/routing/v1/*`)
- **Response**: Routing decision endpoint returns `session_id` and `pinned: true/false` fields
- **Docs**: Added model affinity section to `llm_router.rst`, config reference, and routing API docs
- **Demo**: Model affinity examples in `model_routing_service/` demo (examples 7-9)

### Config example

```yaml
routing:
  session_ttl_seconds: 600
  session_max_entries: 10000
```

### Client usage

```python
from openai import OpenAI
import uuid

client = OpenAI(base_url="http://localhost:12000/v1", api_key="EMPTY")
affinity_id = str(uuid.uuid4())

response = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=messages,
    extra_headers={"X-Model-Affinity": affinity_id},
)
```

### DIFF
--- a/docs/source/resources/includes/plano_config_full_reference_rendered.yaml
+++ b/docs/source/resources/includes/plano_config_full_reference_rendered.yaml
@@ -215,6 +215,9 @@ ratelimits:
   selector:
     key: x-org-id
     value: acme-corp
+routing:
+  session_max_entries: 10000
+  session_ttl_seconds: 600
 state_storage:
   type: memory
 system_prompt: 'You are a helpful assistant. Always respond concisely and accurately.


salmanap:
The test case that would prove this is an agentic loop. Can you build some test cases for that and test with live clients like anthropic and/or openai?

@nehcgs please review. The idea is that if the model is not done, we should pin the agentic loop to a single model until we get the [done] flag from the model, at which point we can resume our model routing logic.

demos/llm_routing/session_pinning/config.yaml
```diff
@@ -0,0 +1,27 @@
+version: v0.3.0
+
+listeners:
+  - type: model
+    name: model_listener
+    port: 12000
+
+model_providers:
```
salmanap: Can you please show how the TTL settings will show up here?

PR #879, opened 2026-04-09
Redis-backed session cache for cross-replica model affinity

Spherrrical:
Introduces a `SessionCache` trait with two backends: the existing in-memory store (default, no config change needed) and a new Redis backend that enables session affinity to persist across multiple brightstaff replicas. The backend is selected via a new `routing.session_cache` config block (`type: memory|redis`, `url`). Includes two end-to-end demos — a Docker Compose quickstart and a full Kubernetes deployment with manifests — each with a verification script that asserts model pinning holds across rounds.


Spherrrical:
Tests conducted show success state across Docker deployment and K8 deployments.

image

image

config/plano_config_schema.yaml
```diff
@@ -441,6 +441,20 @@ properties:
...
+            type: string
+            enum:
+              - memory
+              - redis
+            default: memory
+            description: Session cache backend. "memory" (default) is in-process; "redis" is shared across replicas.
+          url:
+            type: string
+            description: Redis URL, e.g. redis://localhost:6379. Required when type is redis.
+        additionalProperties: false
```
adilhafeez, line 462: It will be deployed to a multi-tenant env so we need a way to store tenant id in the key somewhere. Also prefix should have some namespace like plano:affinity:{tenant_id}:{session_id}

crates/brightstaff/src/router/llm.rs
```diff
@@ -104,65 +93,49 @@ impl RouterService {
...
-        if removed > 0 {
-            info!(
-                removed = removed,
-                remaining = cache.len(),
-                "cleaned up expired session cache entries"
-            );
-        }
+        self.session_cache.cleanup_expired().await;
+    }
+
```
adilhafeez: only memory based cache would very need this method - move it to memory base implementation since redis based cache would handle expired entires by itself

crates/brightstaff/src/session_cache/memory.rs
```diff
@@ -0,0 +1,73 @@
...
+                .min_by_key(|(_, (_, inserted_at))| *inserted_at)
+                .map(|(k, _)| k.clone())
+            {
+                store.remove(&oldest_key);
+            }
+        }
+        store.insert(session_id.to_string(), (route, Instant::now()));
+    }
+
+    async fn remove(&self, session_id: &str) {
```
adilhafeez: right now insert is quite inefficient - I think you're looking for https://docs.rs/lru/latest/lru/ so you can remove oldest and insert in o(1)

crates/brightstaff/src/session_cache/memory.rs
```diff
@@ -0,0 +1,73 @@
...
+        let removed = before - store.len();
+        if removed > 0 {
+            info!(
+                removed = removed,
+                remaining = store.len(),
+                "cleaned up expired session cache entries"
+            );
+        }
+    }
+}
```
adilhafeez: as mentioned earlier in the code review - handle key expiry through an background job here in this file

crates/brightstaff/src/session_cache/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+
+    /// Store a routing decision in the session cache with the given TTL.
+    async fn put(&self, session_id: &str, route: CachedRoute, ttl: Duration);
+
+    /// Remove a cached routing decision by session ID.
+    async fn remove(&self, session_id: &str);
+
+    /// Remove all expired entries. No-op for backends that handle expiry natively (e.g. Redis).
+    async fn cleanup_expired(&self);
+}
```
adilhafeez: this shouldn't be part of standard trait

crates/brightstaff/src/main.rs
```diff
@@ -361,6 +363,54 @@ async fn init_app_state(
     })
 }
 
+/// Initialize the session cache backend based on config.
+/// Defaults to the in-memory backend when no `session_cache` is configured.
+async fn init_session_cache(
+    config: &Configuration,
```
adilhafeez: move this to separate file so to keep main.rs simple

crates/brightstaff/src/session_cache/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+pub trait SessionCache: Send + Sync {
+    /// Look up a cached routing decision by session ID.
+    async fn get(&self, session_id: &str) -> Option<CachedRoute>;
+
+    /// Store a routing decision in the session cache with the given TTL.
+    async fn put(&self, session_id: &str, route: CachedRoute, ttl: Duration);
+
+    /// Remove a cached routing decision by session ID.
+    async fn remove(&self, session_id: &str);
+
```
adilhafeez: let's rename session_id to key as its more generic

adilhafeez (changes requested):
let's not create two separate demos - let's rather update docs and explain how to use this feature

adilhafeez (commented):
pls update rst files and explain how to use redis cache in docs

adilhafeez (commented):
log_affinity_hit is this used anywhere?

and tenant_id is not getting set in redis key - we should do that. One way to pass that could be thorugh config for example,


```
routing:
  session_ttl_seconds: 600
  session_cache:
    type: redis
    url: redis://localhost:6379
    tenant_header: x-org-id       # optional; when set, value is used in cache key
```

and if tenant_id id set then key woud look like htis,

plano:affinity:{tenant_id}:{key}

adilhafeez (approved):

adilhafeez:
pls remove demos before you merge

PR #881, opened 2026-04-10
Add first-class ChatGPT subscription provider support

Spherrrical:
Adds support for routing LLM traffic through a ChatGPT Plus/Pro subscription using OpenAI's device-code OAuth flow (same as the Codex CLI). Users authenticate once via `planoai chatgpt login` and tokens are auto-refreshed from `~/.plano/chatgpt/auth.json`. The `chatgpt` provider is wired end-to-end: CLI commands, config auto-injection, schema updates, Rust `ProviderId`/`LlmProviderType` variants, request normalization for the Codex backend, and custom header forwarding in the WASM gateway. A self-contained demo with `config.yaml`, `chat.py`, and a curl test script is included under `demos/llm_routing/chatgpt_subscription/`.


Stoff81:
Testing out now

Stoff81:
Can confirm this is working for me! Lovely stuff

crates/hermesllm/src/providers/request.rs
```diff
@@ -89,6 +89,41 @@ impl ProviderRequestType {
...
+                        req.instructions =
+                            Some(format!("{}\n\n{}", CHATGPT_BASE_INSTRUCTIONS, existing));
+                    }
+                    None => {
+                        req.instructions = Some(CHATGPT_BASE_INSTRUCTIONS.to_string());
+                    }
+                }
+                req.store = Some(false);
+                req.stream = Some(true);
+
```
adilhafeez, line 114: why are hard-coding system prompt? and what if user has set stream=false?

cli/planoai/config_generator.py
```diff
@@ -28,8 +28,13 @@
     "xai",
     "moonshotai",
     "zhipu",
+    "chatgpt",
 ]
 
+CHATGPT_API_BASE = "https://chatgpt.com/backend-api/codex"
+CHATGPT_DEFAULT_ORIGINATOR = "codex_cli_rs"
+CHATGPT_DEFAULT_USER_AGENT = "codex_cli_rs/0.0.0 (Unknown 0; unknown) unknown"
+
```
adilhafeez, line 38: Why harcode user agent and orignator? This will overwrite user's user_agent if it was set.

cli/planoai/config_generator.py
```diff
@@ -331,6 +336,25 @@ def validate_and_render_schema():
...
                 del model_provider["provider"]
+
+            # Auto-wire ChatGPT provider: inject base_url, access_key, and extra headers
+            if provider == "chatgpt":
+                if not model_provider.get("base_url"):
+                    model_provider["base_url"] = CHATGPT_API_BASE
+                if not model_provider.get("access_key"):
+                    model_provider["access_key"] = "$CHATGPT_ACCESS_TOKEN"
+                import uuid
+
```
adilhafeez: move this import to top

cli/planoai/main.py
```diff
@@ -118,6 +119,33 @@ def _temporary_cli_log_level(level: str | None):
...
 
 
+def _inject_chatgpt_tokens_if_needed(plano_config_file, env, console):
+    """If config uses chatgpt providers, resolve tokens from ~/.plano/chatgpt/auth.json."""
+    import yaml
+
+    with open(plano_config_file, "r") as f:
+        config = yaml.safe_load(f)
+
+    providers = config.get("model_providers") or config.get("llm_providers") or []
```
adilhafeez: yaml is already parsed in `up` method - reuse that object instead of parsing it again

adilhafeez (changes requested):
the refresh token flow is not automatic - so the way it works right now is that it 1. loads token from disk, 2. refresh token if needed and then 3. inject the token. Issue is if token expired while plano is running there is no way to refersh it and we'll have to restart the service to refresh the token. This doesn't seem right.

crates/hermesllm/src/providers/request.rs
Spherrrical: the system will not work without this prompt, and the stream=false does not work for this provider

cli/planoai/config_generator.py
Spherrrical: the provider will not work without these headers

Stoff81:
I have found an issue with using OpenClaw and this branch, when posting with tools as OpenClaw does:

`curl -sS http://localhost:12000/v1/responses \
  -H "Content-Type: application/json" \
  -d '{
    "model": "gpt-5.3-codex",
    "input": [
      {
        "type": "message",
        "role": "user",
        "content": [
          { "type": "input_text", "text": "What is the capital of France?" }
        ]
      }
    ],
    "stream": false,
    "tools": [
      {
        "type": "function",
        "name": "get_time",
        "description": "Get the current time",
        "parameters": {
          "type": "object",
          "properties": {},
          "additionalProperties": false
        }
      }
    ],
    "tool_choice": "auto"
  }'`
  
  I am getting `Failed to parse request: missing field `name` at line 1 column 330%`
  
  I have created a PR to address this here: https://github.com/katanemo/plano/pull/884

Stoff81:
I found an issue with using OpenClaw and this branch, when posting:


`{
  "model": "gpt-5.3-codex",
  "input": [
    {
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "what model are you on"
        }
      ]
    }
  ],
  "include": [
    "reasoning.encrypted_content"
  ],
  "store": false,
  "instructions": "You are Codex, based on GPT-5. You are running as a coding agent in the Codex CLI on a user's computer.",
  "stream": true,
  "max_output_tokens": 8192
}`

We were getting Unsupported Field: max_output_tokens

I have created a PR to address this here: https://github.com/katanemo/plano/pull/884

Stoff81:
I found an issue with using OpenClaw and this branch, when posting:

`{ "model": "gpt-5.3-codex", "input": [ { "role": "user", "content": [ { "type": "input_text", "text": "what model are you on" } ] } ], "include": [ "reasoning.encrypted_content" ], "store": false, "instructions": "You are Codex, based on GPT-5. You are running as a coding agent in the Codex CLI on a user's computer.", "stream": true, "max_output_tokens": 8192 }`

We were getting 

`{
  "error": {
    "message": "Invalid value: 'input_text'. Supported values are: 'output_text' and 'refusal'.",
    "type": "invalid_request_error",
    "param": "input[12].content[0]",
    "code": "invalid_value"
  }
}%`

I have created a PR to address this here: https://github.com/katanemo/plano/pull/884

demos/llm_routing/chatgpt_subscription/README.md
```diff
@@ -0,0 +1,61 @@
+# ChatGPT Subscription Routing
+
+Route requests through your ChatGPT Plus/Pro subscription using Plano. Uses the OpenAI Responses API under the hood, targeting `chatgpt.com/backend-api/codex/responses`.
+
+## Setup
+
+### 1. Authenticate with ChatGPT
+
+```bash
+planoai chatgpt login
```
adilhafeez, line 10: can we follow similar pattern like claude /login?

adilhafeez (changes requested):
I don't like the concept of adding watchdog and restarting the entire process. We could use passthrough to get auth token from client. Or use already stored client_id/client_secret to refresh the token from within brightstaff.

demos/llm_routing/chatgpt_subscription/README.md
Spherrrical: this is a better way to do it - in the event we add support for other providers (ex. planoai gemini login) or etc. It's easy to use

demos/llm_routing/chatgpt_subscription/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+This opens a device code flow — visit the URL shown and enter the code. Tokens are saved to `~/.plano/chatgpt/auth.json`.
+
+### 2. Start Plano
```
adilhafeez, line 15: try without config and see if it works https://github.com/katanemo/plano/blob/main/cli/planoai/defaults.py#L39-L83

adilhafeez (approved):
conditionally approve - but first need to figure out model name collision

PR #889, opened 2026-04-17
Add DigitalOcean as a first-class LLM provider

adilhafeez:
## Summary
- Adds `ProviderId::DigitalOcean` (canonical key `digitalocean`, aliases `do` / `do_ai`). DO's Inference endpoint is OpenAI-compatible; reuses the shared OpenAI chat-completions path in hermesllm.
- Ships DO's catalog in `provider_models.yaml` under a `digitalocean:` key: concrete OpenAI/Anthropic/open-weight models hosted on DO, plus the `router:software-engineering` alias.
- Makes `digitalocean` a **first-class provider** in config_generator: adds it to `SUPPORTED_PROVIDERS_WITHOUT_BASE_URL` so users can write `model: digitalocean/openai-gpt-5.4` without supplying `base_url` / `provider_interface`.
- Adds the `digitalocean` cluster in `envoy.template.yaml` pointing at `inference.do-ai.run:443` with TLS/SNI, matching the pattern used for `openai`, `anthropic`, etc.

## User-visible result
```yaml
model_providers:
  - model: digitalocean/openai-gpt-5.4
    access_key: $DO_API_KEY
```
validates, renders, and routes to `https://inference.do-ai.run/v1/chat/completions` — no `base_url` required.

## Test plan
- [ ] `cd crates && cargo test --lib -p hermesllm` (145 existing tests still pass)
- [ ] `cargo clippy --locked -p hermesllm --all-targets -- -D warnings`
- [ ] `cd cli && uv run pytest test/test_config_generator.py` (10 tests)
- [ ] Smoke: `planoai up` with a minimal `digitalocean/` config and send a chat/completions request against `localhost:12000`

PR #890, opened 2026-04-17
Zero-config planoai up: pass-through proxy with auto-detected providers

adilhafeez:
## Summary
- `planoai up` now works with no config file: synthesizes a pass-through proxy covering the common LLM providers (OpenAI, Anthropic, Gemini, Groq, DeepSeek, Mistral, DO) on the default listener port `:12000`.
- For providers with their env var set (`OPENAI_API_KEY`, etc.), Plano binds the key via `access_key: \$ENV_VAR`. For providers without env keys, it sets `passthrough_auth: true` so the client's own `Authorization` header is forwarded. No env vars are required to start.
- Tracing is auto-wired to `http://localhost:4317` with full sampling, so observability tooling (`planoai trace`, planned `planoai obs`) works out of the box.
- Synthesized config is written to `~/.plano/default_config.yaml` and handed through the existing validate+render pipeline unchanged; explicit user-provided configs still take precedence.

## Test plan
- [ ] `cd cli && uv run pytest test/test_defaults.py` (6 new tests cover env detection, schema validity, tracing block, provider registry)
- [ ] `uv run pytest test/` — full existing suite passes
- [ ] Smoke: `cd /tmp/empty && planoai up` (no env keys) → proxy starts, prints `pass-through: openai, anthropic, gemini, groq, deepseek, mistral, do`
- [ ] Smoke: `export OPENAI_API_KEY=sk-...; planoai up` → prints `env-keyed: openai | pass-through: anthropic, ...`
- [ ] Smoke: `curl localhost:12000/v1/chat/completions -H "Authorization: Bearer \$KEY" -d '{"model":"openai/gpt-4o-mini","messages":[...]}'` succeeds end-to-end

### DIFF
--- a/cli/planoai/defaults.py
+++ b/cli/planoai/defaults.py
@@ -72,14 +72,13 @@ PROVIDER_DEFAULTS: list[ProviderDefault] = [
         base_url="https://api.mistral.ai/v1",
         model_pattern="mistral/*",
     ),
-    # DigitalOcean Gradient is OpenAI-compatible and `do` is not in the built-in
-    # SUPPORTED_PROVIDERS list, so we must provide provider_interface explicitly.
+    # DigitalOcean Gradient is now a first-class provider (see #889) — model
+    # prefix `digitalocean/` routes to the built-in cluster, no base_url needed.
     ProviderDefault(
-        name="do",
+        name="digitalocean",
         env_var="DO_API_KEY",
         base_url="https://inference.do-ai.run/v1",
-        model_pattern="do/*",
-        provider_interface="openai",
+        model_pattern="digitalocean/*",
     ),
 ]
 
--- a/cli/planoai/defaults.py
+++ b/cli/planoai/defaults.py
@@ -1,0 +1,167 @@
+"""Default config synthesizer for zero-config ``planoai up``.
+
+When the user runs ``planoai up`` in a directory with no ``config.yaml`` /
+``plano_config.yaml``, we synthesize a pass-through config that covers the
+common LLM providers and auto-wires OTel export to ``localhost:4317`` so
+``planoai obs`` works out of the box.
+
+Auth handling:
+- If the provider's env var is set, bind ``access_key: $ENV_VAR``.
+- Otherwise set ``passthrough_auth: true`` so the client's own Authorization
+  header is forwarded. No env var is required to start the proxy.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+
+
+DEFAULT_LLM_LISTENER_PORT = 12000
+# plano_config validation requires an http:// scheme on the OTLP endpoint.
+DEFAULT_OTLP_ENDPOINT = "http://localhost:4317"
+
+
+@dataclass(frozen=True)
+class ProviderDefault:
+    name: str
+    env_var: str
+    base_url: str
+    model_pattern: str
+    # Only set for providers whose prefix in the model pattern is NOT one of the
+    # built-in SUPPORTED_PROVIDERS in cli/planoai/config_generator.py. For
+    # built-ins, the validator infers the interface from the model prefix and
+    # rejects configs that set this field explicitly.
+    provider_interface: str | None = None
+
+
+# Keep ordering stable so synthesized configs diff cleanly across runs.
+PROVIDER_DEFAULTS: list[ProviderDefault] = [
+    ProviderDefault(
+        name="openai",
+        env_var="OPENAI_API_KEY",
+        base_url="https://api.openai.com/v1",
+        model_pattern="openai/*",
+    ),
+    ProviderDefault(
+        name="anthropic",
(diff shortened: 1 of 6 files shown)


cli/planoai/defaults.py
```diff
@@ -0,0 +1,162 @@
...
+        row.update(base)
+        return row
+
+    model_providers: list[dict] = []
+    for p in detection.with_keys:
+        model_providers.append(_entry(p, {"access_key": f"${p.env_var}"}))
+    for p in detection.passthrough:
+        model_providers.append(_entry(p, {"passthrough_auth": True}))
+
+    # We intentionally don't mark any provider as `default: true` here:
```
Spherrrical: Doesnt at least model need to be marked as default?

cli/planoai/main.py
```diff
@@ -328,12 +328,30 @@ def up(file, path, foreground, with_tracing, tracing_port, docker, verbose):
         # Use the utility function to find config file
         plano_config_file = find_config_file(path, file)
 
-        # Check if the file exists
+        # Zero-config fallback: when no user config is present, synthesize a
+        # pass-through config that covers the common LLM providers and
+        # auto-wires OTel export to ``planoai obs``. See cli/planoai/defaults.py.
         if not os.path.exists(plano_config_file):
+            import yaml
```
Spherrrical: move imports to the top

cli/planoai/defaults.py
```diff
@@ -0,0 +1,162 @@
...
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass
+
+DEFAULT_LLM_LISTENER_PORT = 12000
+# plano_config validation requires an http:// scheme on the OTLP endpoint.
+DEFAULT_OTLP_ENDPOINT = "http://localhost:4317"
```
Spherrrical, line 21: can we change this value with a flag on `up`?

cli/planoai/defaults.py
```diff
@@ -0,0 +1,162 @@
...
+        model_pattern="deepseek/*",
+    ),
+    ProviderDefault(
+        name="mistral",
+        env_var="MISTRAL_API_KEY",
+        base_url="https://api.mistral.ai/v1",
+        model_pattern="mistral/*",
+    ),
+    # DigitalOcean Gradient is OpenAI-compatible and `do` is not in the built-in
+    # SUPPORTED_PROVIDERS list, so we must provide provider_interface explicitly.
```
Spherrrical: #889 solves the `SUPPORTED_PROVIDERS` list for digitalocean, let's make that merge first then refactor this to include that change

Spherrrical:
One more thing - are we checking if we have at least one of the 5-6 providers on default? Ex. if I dont have `anthropic` but i have an `openai` key, will `planoai up` still succeed?

cli/planoai/defaults.py
adilhafeez: its done in 889

cli/planoai/defaults.py
adilhafeez: updated

adilhafeez:
Thanks for the review @Spherrrical — addressed in the latest push:

1. **Default model** — when ≥1 env-keyed provider is detected, we now promote a concrete model (`openai/gpt-4o-mini` / `anthropic/claude-haiku-4-5` / `digitalocean/openai-gpt-5.4-mini`, whichever is first in `PROVIDER_DEFAULTS`) to `default: true`. Lets bare-name requests route somewhere sensible. When no env keys are set we stay in pure pass-through (no default) — clients must prefix the model name.
2. **Imports at top** — `yaml` and the `planoai.defaults` imports moved to the top of `main.py`.
3. **Port flag** — added `--listener-port` on `planoai up` (default 12000 from `DEFAULT_LLM_LISTENER_PORT`). Ignored when a config file is present.
4. **#889 rebase** — merged in and dropped the now-redundant `provider_interface="openai"` on the DigitalOcean entry since it's a first-class provider post-#889.

> are we checking if we have at least one of the 5-6 providers on default? Ex. if I dont have `anthropic` but i have an `openai` key, will `planoai up` still succeed?

Yes — `planoai up` succeeds even with zero env keys set. Providers without a key just run in pass-through mode (client must send the `Authorization` header). When you have only `OPENAI_API_KEY`, openai uses the key and the rest are pass-through; the default entry then points at `openai/gpt-4o-mini`.

Spherrrical (approved):

adilhafeez:
Reverted the `default_model` logic — @adilhafeez rightly pointed out the model field is always required, and brightstaff's wildcard expansion already registers bare names as lookup keys (see [llm_providers.rs:219-220](https://github.com/katanemo/plano/blob/main/crates/common/src/llm_providers.rs#L219-L220)). So `{"model": "gpt-4o-mini"}` without a prefix already resolves to openai via the wildcard — no synthesized `default: true` row needed. Simpler config.

PR #891, opened 2026-04-17
planoai obs: live LLM observability TUI

adilhafeez:
## Summary
- New CLI subcommand `planoai obs` — foreground rich TUI that receives OTLP/gRPC spans from brightstaff on :4317, keeps an in-memory ring buffer (default 1000), and renders per-request + aggregate LLM observability. No SQLite, no web UI, no external backend.
- Brightstaff now emits per-request usage on LLM spans: prompt / completion / total tokens, cached input tokens, cache-creation tokens, reasoning tokens, TTFT, duration, status, `plano.session_id` (from `x-model-affinity`), and `plano.route.name` (from Plano routing output *or* the upstream's `x-model-router-selected-route` response header).
- ObservableStreamProcessor buffers response bytes and best-effort-extracts usage + resolved model from both non-streaming JSON bodies and SSE streams, overriding `llm.model` with the model the upstream actually ran (e.g. `router:software-engineering` → `openai-gpt-5.4`) so cost lookup keys off the real model.
- Hermesllm `TokenUsage` trait gets optional `cached_input_tokens()`, `cache_creation_tokens()`, `reasoning_tokens()` methods + a richer `UsageDetails` struct. Implementations added for OpenAI, Anthropic, and OpenAI Responses usage types.
- Cost computation in the CLI fetches DO's Gradient catalog (`api.digitalocean.com/v2/gen-ai/models/catalog`) at startup with `DIGITALOCEAN_TOKEN` / `DO_PAT` / `DO_API_KEY` auth; treats `input_price_per_million` values < 1 as already per-token (matches DO's catalog shape).

## TUI layout
- Header panels: Request (most recent call) / Cost (tokens + $/request) / Totals (count, cost, tokens, sessions)
- Totals by model table with cache write/read split
- Route hit % panel (hidden when no calls carry a route)
- Recent requests scrolling log with adaptive cost formatting

## Test plan
- [ ] `cd cli && uv run pytest test/test_obs_collector.py test/test_obs_pricing.py test/test_obs_render.py` (16 tests)
- [ ] `cd crates && cargo test --lib -p brightstaff -p hermesllm && cargo clippy --locked --all-targets -- -D warnings && cargo fmt --all -- --check`
- [ ] Smoke: in one terminal `planoai obs`; in another `planoai up` with `tracing.opentracing_grpc_endpoint: http://localhost:4317`; send a few chat/completions — verify rows appear with tokens, duration, TTFT, and (if `DIGITALOCEAN_TOKEN` set) cost
- [ ] Smoke: send a DO `router:software-engineering` request — verify `llm.model` is overridden with the resolved concrete model and `plano.route.name` reflects DO's classification header

### DIFF
--- a/cli/planoai/obs/collector.py
+++ b/cli/planoai/obs/collector.py
@@ -1,0 +1,281 @@
+"""In-memory collector for LLM calls, fed by OTLP/gRPC spans from brightstaff."""
+
+from __future__ import annotations
+
+import threading
+from collections import deque
+from concurrent import futures
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import Any, Iterable
+
+import grpc
+from opentelemetry.proto.collector.trace.v1 import (
+    trace_service_pb2,
+    trace_service_pb2_grpc,
+)
+
+
+DEFAULT_GRPC_PORT = 4317
+DEFAULT_CAPACITY = 1000
+
+
+@dataclass
+class LLMCall:
+    """One LLM call as reconstructed from a brightstaff LLM span.
+
+    Fields default to ``None`` when the underlying span attribute was absent.
+    """
+
+    request_id: str
+    timestamp: datetime
+    model: str
+    provider: str | None = None
+    request_model: str | None = None
+    session_id: str | None = None
+    route_name: str | None = None
+    is_streaming: bool | None = None
+    status_code: int | None = None
+    prompt_tokens: int | None = None
+    completion_tokens: int | None = None
+    total_tokens: int | None = None
+    cached_input_tokens: int | None = None
+    cache_creation_tokens: int | None = None
+    reasoning_tokens: int | None = None
+    ttft_ms: float | None = None
+    duration_ms: float | None = None
+    routing_strategy: str | None = None
+    routing_reason: str | None = None
+    cost_usd: float | None = None
+
+    @property
+    def tpt_ms(self) -> float | None:
+        if self.duration_ms is None or self.completion_tokens in (None, 0):
+            return None
+        ttft = self.ttft_ms or 0.0
+        generate_ms = max(0.0, self.duration_ms - ttft)
+        if generate_ms <= 0:
+            return None
+        return generate_ms / self.completion_tokens
+
+    @property
+    def tokens_per_sec(self) -> float | None:
+        tpt = self.tpt_ms
+        if tpt is None or tpt <= 0:
+            return None
+        return 1000.0 / tpt
+
+
+class LLMCallStore:
+    """Thread-safe ring buffer of recent LLM calls."""
+
+    def __init__(self, capacity: int = DEFAULT_CAPACITY) -> None:
+        self._capacity = capacity
+        self._calls: deque[LLMCall] = deque(maxlen=capacity)
+        self._lock = threading.Lock()
+
+    @property
+    def capacity(self) -> int:
+        return self._capacity
+
+    def add(self, call: LLMCall) -> None:
+        with self._lock:
+            self._calls.append(call)
+
+    def clear(self) -> None:
+        with self._lock:
+            self._calls.clear()
+
+    def snapshot(self) -> list[LLMCall]:
+        with self._lock:
+            return list(self._calls)
+
+    def __len__(self) -> int:
+        with self._lock:
+            return len(self._calls)
+
+
+# Attribute keys mirror crates/brightstaff/src/tracing/constants.rs.
+_LLM_MODEL = "llm.model"
+_LLM_PROVIDER = "llm.provider"
+_LLM_IS_STREAMING = "llm.is_streaming"
+_LLM_DURATION_MS = "llm.duration_ms"
+_LLM_TTFT_MS = "llm.time_to_first_token"
+_LLM_PROMPT_TOKENS = "llm.usage.prompt_tokens"
+_LLM_COMPLETION_TOKENS = "llm.usage.completion_tokens"
+_LLM_TOTAL_TOKENS = "llm.usage.total_tokens"
+_LLM_CACHED_INPUT_TOKENS = "llm.usage.cached_input_tokens"
+_LLM_CACHE_CREATION_TOKENS = "llm.usage.cache_creation_tokens"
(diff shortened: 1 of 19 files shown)


cli/planoai/obs/collector.py
```diff
@@ -0,0 +1,280 @@
...
+        with self._lock:
+            return list(self._calls)
+
+    def __len__(self) -> int:
+        with self._lock:
+            return len(self._calls)
+
+
+# Attribute keys mirror crates/brightstaff/src/tracing/constants.rs.
+_LLM_MODEL = "llm.model"
```
Spherrrical: Instead of mapping each otel attribute to a const, are we able to directly just use the otel attribute? if not its fine this code just feels a little redundant in that sense

cli/planoai/obs/collector.py
```diff
@@ -0,0 +1,280 @@
...
+        server = grpc.server(futures.ThreadPoolExecutor(max_workers=4))
+        trace_service_pb2_grpc.add_TraceServiceServicer_to_server(
+            _ObsServicer(self.store, self.pricing), server
+        )
+        address = f"{self.host}:{self.port}"
+        bound = server.add_insecure_port(address)
+        if bound == 0:
+            raise OSError(
+                f"Failed to bind OTLP listener on {address}: port already in use. "
+                "Stop `planoai trace listen` or pick another port with --port."
```
Spherrrical: a little clearer here can be:

`"Stop tracing via `planoai trace down` or pick another port with --port."

cli/planoai/obs/pricing.py
```diff
@@ -0,0 +1,275 @@
+"""DigitalOcean Gradient pricing catalog for the obs console.
```
Spherrrical, line 1: What if someone wants to use a different pricing structure - like Vercel AI Gateway or OpenRouter?

cli/planoai/obs/pricing.py
```diff
@@ -0,0 +1,275 @@
...
+    @classmethod
+    def fetch(
+        cls,
+        url: str = DEFAULT_PRICING_URL,
+        api_key: str | None = None,
+    ) -> "PricingCatalog":
+        """Fetch pricing from DO's catalog endpoint. On failure, returns an
+        empty catalog (cost column will be blank).
+
+        The catalog endpoint requires a DigitalOcean Personal Access Token —
```
Spherrrical: the experience of running `planoai obs` should feel like a good run experience. Although I understand that it requires a token, it might feel broken to the first run experience requiring a user to signup externally.

docs/source/get_started/quickstart.rst
```diff
@@ -340,6 +340,42 @@ And to get the list of supported currencies:
    "Here is a list of the currencies that are supported for conversion from USD, along with their symbols:\n\n1. AUD - Australian Dollar\n2. BGN - Bulgarian Lev\n3. BRL - Brazilian Real\n4. CAD - Canadian Dollar\n5. CHF - Swiss Franc\n6. CNY - Chinese Renminbi Yuan\n7. CZK - Czech Koruna\n8. DKK - Danish Krone\n9. EUR - Euro\n10. GBP - British Pound\n11. HKD - Hong Kong Dollar\n12. HUF - Hungarian Forint\n13. IDR - Indonesian Rupiah\n14. ILS - Israeli New Sheqel\n15. INR - Indian Rupee\n16. ISK - Icelandic Króna\n17. JPY - Japanese Yen\n18. KRW - South Korean Won\n19. MXN - Mexican Peso\n20. MYR - Malaysian Ringgit\n21. NOK - Norwegian Krone\n22. NZD - New Zealand Dollar\n23. PHP - Philippine Peso\n24. PLN - Polish Złoty\n25. RON - Romanian Leu\n26. SEK - Swedish Krona\n27. SGD - Singapore Dollar\n28. THB - Thai Baht\n29. TRY - Turkish Lira\n30. USD - United States Dollar\n31. ZAR - South African Rand\n\nIf you want to convert USD to any of these currencies, you can select the one you are interested in."
 
 
+Observability Console
```
Spherrrical: talk about observability but provide 2 options - one being this `obs` command but also tracing via `trace`. That way theres 2 routes to see cost tracking vs the line of requests etc.

Spherrrical (changes requested):

cli/planoai/obs/collector.py
adilhafeez: right we don't need those consts can just use names directly

adilhafeez: updated code

cli/planoai/obs/collector.py
adilhafeez: fixed

cli/planoai/obs/pricing.py
adilhafeez: I'll like to keep it as DO for now we can add support for generic pricing in follow up PR

cli/planoai/obs/pricing.py
adilhafeez: it doesn't need token - I've updated the comment

adilhafeez:
Thanks for the review @Spherrrical — addressed:

1. **`collector.py:98` — drop OTel attr consts**: inlined all 19 attribute strings directly in `span_to_llm_call`. ~20 lines lighter and easier to read.
2. **`collector.py:272` — port-in-use wording**: updated to "Stop tracing via `planoai trace down` or pick another port with --port."
3. **`pricing.py` pluggability (Vercel AI Gateway / OpenRouter)**: leaving as DO-only for this PR; will tackle in a follow-up that introduces a `PricingProvider` abstraction so adding OpenRouter / Vercel / static-yaml is clean.
4. **`pricing.py:61` — first-run UX**: my docstring was wrong — DO's pricing catalog (`api.digitalocean.com/v2/gen-ai/models/catalog`) is actually a **public endpoint**, no PAT or signup required. Stripped all auth logic from `PricingCatalog.fetch`, updated docstring + obs startup message. Cost populates on first run out of the box.
5. **`quickstart.rst:343` — obs + trace docs**: restructured the Observability section around the two complementary tools, with a table showing when to use each, and an example ("spot p95 spike in obs → jump to trace for the slow request id").

Spherrrical (approved):

PR #895, opened 2026-04-18
fix: truncate oversized user messages in orchestrator routing prompt

adilhafeez:
## Summary

Hardens brightstaff's orchestrator routing path against the class of production incident where a single large user message blew through the upstream model's context window and surfaced as a confusing parse error. Three layered fixes:

1. **Trim overflowing user messages** instead of letting them pass through whole. Uses a middle-trim (head + ellipsis + tail) so both the task framing and the user's actual ask survive truncation.
2. **Cap the routing prompt at the last 16 filtered turns** as an outer guardrail, so prompt growth is predictable on long-running sessions regardless of per-message size.
3. **Parse upstream error responses** instead of deserializing every response as a success, so a 400 "context length exceeded" actually shows up as that in logs and errors rather than `missing field 'id'`.

### DIFF
--- a/crates/brightstaff/src/router/http.rs
+++ b/crates/brightstaff/src/router/http.rs
@@ -1,7 +1,13 @@
 use hermesllm::apis::openai::ChatCompletionsResponse;
 use hyper::header;
+use serde::Deserialize;
 use thiserror::Error;
 use tracing::warn;
+
+/// Max bytes of raw upstream body we include in a log message or error text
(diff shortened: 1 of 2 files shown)


salmanap (approved):
Please fix the reversing of the messages early in the code path. Else you will have the first 16 max messages.

ahafeez-do:
> Please fix the reversing of the messages early in the code path. Else you will have the first 16 max messages.

@salmanap verified and its working as expected

PR #898, opened 2026-04-18
fix(anthropic-stream): avoid bare/duplicate message_stop on OpenAI upstream

adilhafeez:
## Summary

Fixes the intermittent \`Received message_stop without a current message\` error seen in Claude Code (and other strict Anthropic SSE clients) when routed through plano against an OpenAI-compatible upstream.

Three overlapping problems in \`AnthropicMessagesStreamBuffer\` combined to produce bare or duplicate \`message_stop\` events on the wire:

1. **Bare \`message_stop\`**: the \`MessageStop\` handler pushed the event unconditionally, so an upstream stream that went straight to \`[DONE]\` without any content (empty/filtered/errored completions) emitted \`message_stop\` with no preceding \`message_start\`.
2. **Double \`message_stop\`**: \`to_bytes()\` synthesizes a \`message_stop\` whenever \`seen_message_delta\` is true. Since \`to_bytes()\` is called once per HTTP body chunk ([llm_gateway/src/stream_context.rs:582](crates/llm_gateway/src/stream_context.rs#L582)), a stream whose \`finish_reason\` chunk and \`[DONE]\` marker land in separate HTTP chunks produced two \`message_stop\` frames — explaining the intermittence (and why same-chunk responses were fine).
3. **Missing \`message_start\` before \`message_delta\`**: unlike \`ContentBlockStart\`/\`ContentBlockDelta\`, the \`MessageDelta\` handler did not ensure a \`message_start\` was emitted. A \`finish_reason\`-only response therefore produced \`message_delta\` + synthetic \`message_stop\` with no opener.

## Fix

Turned the flag-bag into a proper state machine guarded by \`message_started\` and a new \`message_stopped\` flag:

- \`ensure_message_started()\` is called by every handler that can legitimately be the first event on the wire.
- \`MessageStop\` now synthesizes whatever envelope pieces are missing (\`message_start\`, optional \`content_block_start\`/\`stop\`, \`message_delta { end_turn }\`) before emitting the terminal frame, so even an empty upstream produces a well-formed envelope.
- Events arriving after \`message_stop\` (stray \`[DONE]\` echoes, late deltas from a racing upstream) are dropped.
- \`to_bytes()\` only synthesizes \`message_stop\` when \`seen_message_delta && !message_stopped\`, eliminating the double-close across chunk boundaries.

All existing tests pass unchanged; 3 new regression tests reproduce each failure mode.

## Test plan

- [x] \`cd crates && cargo fmt --all -- --check\`
- [x] \`cd crates && cargo clippy --locked --all-targets --all-features -- -D warnings\` (full workspace, incl. \`llm_gateway\` / \`brightstaff\`)
- [x] \`cd crates && cargo test --lib -p hermesllm\` (148 passed, incl. 3 new regression tests)
- [ ] Smoke test Claude Code against plano + OpenAI on a mix of empty responses, tool calls, and long streams that fragment across HTTP chunks

### New regression tests

- \`test_openai_to_anthropic_emits_single_message_stop_across_chunk_boundary\` — simulates the intermittent production failure by splitting \`finish_reason\` and \`[DONE]\` into two separate chunks with a \`to_bytes()\` flush in between.
- \`test_openai_done_only_stream_synthesizes_valid_envelope\` — upstream sends only \`[DONE]\`; asserts a full \`message_start → message_delta → message_stop\` envelope.
- \`test_events_after_message_stop_are_dropped\` — a late \`[DONE]\` after a closed stream produces zero bytes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
+++ b/crates/hermesllm/src/apis/streaming_shapes/anthropic_streaming_buffer.rs
@@ -3,6 +3,7 @@ use crate::apis::anthropic::{
 };
 use crate::apis::streaming_shapes::sse::{SseEvent, SseStreamBufferTrait};
 use crate::providers::streaming_response::ProviderStreamResponseType;
+use log::warn;
 use std::collections::HashSet;
 
 /// SSE Stream Buffer for Anthropic Messages API streaming.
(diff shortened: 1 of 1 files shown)


salmanap (approved):
These changes look good to me.  I saw these idiosyncrasies before, but never saw the client blow up. For example if we send message_stop twice the client silently ignore that. But if this makes the implementation more bullet proof, then awesome.

PR #912, opened 2026-04-24
fix(routing): auto-migrate v0.3.0 inline routing_preferences to v0.4.0 top-level

Spherrrical:
## Summary

- `cli/planoai/config_generator.py` now lifts v0.3.0 inline `model_providers[].routing_preferences` into the v0.4.0 top-level `routing_preferences: [{name, description, models: [...]}]` shape, merges same-named preferences across providers into a single entry with all declaring providers collected into `models`, bumps `version` to `v0.4.0` when migration fires, and prints a single deprecation warning. Pre-existing user-defined top-level entries win over inline migrations; wildcard providers with inline preferences still error.
- `config/plano_config_schema.yaml` still accepts the inline shape so v0.3.0 configs validate, but both inline blocks now carry an explicit `description` flagging them deprecated. Brightstaff only ever sees the rendered top-level + v0.4.0 shape, so no Rust changes were needed.
- Docs and skills are canonicalized to the v0.4.0 top-level multi-model form: `docs/source/guides/llm_router.rst`, `docs/source/resources/includes/plano_config_full_reference.yaml` (+ regenerated rendered variant), `docs/source/concepts/llm_providers/supported_providers.rst`, `skills/rules/routing-preferences.md`, and all inline examples in `skills/AGENTS.md`. Each includes a short note explaining that multiple models per preference enable automatic client fallback on 429/5xx.
- Demo configs under `demos/llm_routing/*` intentionally stay on v0.3.0 with inline preferences (per product direction) and start working again via auto-migration at compile time — the Claude Code router config picks up `claude-sonnet-4-6` as default in the same commit.

## Why

The schema previously permitted both inline (v0.3.0) and top-level (v0.4.0) `routing_preferences`, but brightstaff only deserializes the top-level shape and rejects top-level entries on configs with `version < v0.4.0`. The CLI never bridged the two, so every existing v0.3.0 demo (Claude Code, Codex, preference_based_routing, currency_exchange) silently lost its routing_preferences at runtime. This PR closes that gap without forcing any existing user to edit their configs.

## Test plan

- [x] `cd cli && uv run pytest -v` — 96/96 pass (6 new migration tests: distinct names, same-name merge, user top-level wins, v0.4.0 passthrough, wildcard rejection, no-op on empty)
- [x] Regenerated `plano_config_full_reference_rendered.yaml` by running the CLI against the updated reference — top-level `routing_preferences` with `models: [...]` and `version: v0.4.0`
- [x] Verified `demos/llm_routing/{claude_code_router,codex_router,preference_based_routing,model_routing_service/config.yaml}` + `demos/advanced/currency_exchange/config.yaml` all compile cleanly; v0.3.0 ones emit the deprecation warning and produce top-level `routing_preferences`, v0.4.0 ones pass through untouched
- [x] `planoai up demos/llm_routing/claude_code_router/config.yaml` now reports `Configuration valid` (previously broken end-to-end)
- [ ] Optional follow-up PR to migrate demo YAMLs to the new canonical shape once ready to surface it

### DIFF
--- a/cli/planoai/config_generator.py
+++ b/cli/planoai/config_generator.py
@@ -58,6 +58,104 @@ def get_endpoint_and_port(endpoint, protocol):
         return endpoint, port
 
 
+def migrate_inline_routing_preferences(config_yaml):
+    """Lift v0.3.0-style inline ``routing_preferences`` under each
+    ``model_providers`` entry to the v0.4.0 top-level ``routing_preferences``
+    list with ``models: [...]``.
+
+    Preferences with the same ``name`` across multiple providers are merged
+    into a single top-level entry whose ``models`` list contains every
+    provider's full ``<provider>/<model>`` string in declaration order. The
+    first ``description`` encountered wins; conflicts are warned, not
+    errored, so existing v0.3.0 configs keep compiling. Any top-level
+    preference already defined by the user is preserved as-is.
+
+    Also bumps ``version`` to ``v0.4.0`` when migration produced any entry,
+    so brightstaff's v0.4.0 gate for top-level ``routing_preferences``
+    accepts the rendered config.
+    """
+    model_providers = config_yaml.get("model_providers") or []
+    if not model_providers:
+        return
+
+    migrated = {}
+    for model_provider in model_providers:
+        inline_prefs = model_provider.get("routing_preferences")
+        if not inline_prefs:
+            continue
+
+        full_model_name = model_provider.get("model")
+        if not full_model_name:
+            continue
+
+        if "/" in full_model_name and full_model_name.split("/")[-1].strip() == "*":
+            raise Exception(
+                f"Model {full_model_name} has routing_preferences but uses wildcard (*). Models with routing preferences cannot be wildcards."
+            )
+
+        for pref in inline_prefs:
+            name = pref.get("name")
+            description = pref.get("description", "")
+            if not name:
+                continue
+            if name in migrated:
+                entry = migrated[name]
+                if description and description != entry["description"]:
+                    print(
+                        f"WARNING: routing preference '{name}' has conflicting descriptions across providers; keeping the first one."
+                    )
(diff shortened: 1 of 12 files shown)


skills/AGENTS.md
adilhafeez, line 1: we need to consolidate - we have claude.md and skills in .claude file

cli/planoai/config_generator.py
```diff
@@ -58,6 +58,104 @@ def get_endpoint_and_port(endpoint, protocol):
         return endpoint, port
 
 
+def migrate_inline_routing_preferences(config_yaml):
```
adilhafeez, line 61: this should only work for version < v0.4.0

skills/AGENTS.md
Spherrrical: This AGENTS.md is only for when installing skills - the primary claude md will remain

cli/planoai/config_generator.py
Spherrrical: Good catch, updating

cli/planoai/config_generator.py
```diff
@@ -58,6 +58,104 @@ def get_endpoint_and_port(endpoint, protocol):
...
+    merged = list(existing_top_level)
+    for name, entry in migrated.items():
+        if name in existing_names:
+            continue
+        merged.append(entry)
+    config_yaml["routing_preferences"] = merged
+
+    current_version = str(config_yaml.get("version", ""))
+    if _version_tuple(current_version) < (0, 4, 0):
+        config_yaml["version"] = "v0.4.0"
```
adilhafeez: this check must be at the top

adilhafeez (changes requested):

cli/planoai/config_generator.py
Spherrrical: Updating

adilhafeez (approved):

PR #946, opened 2026-05-17
Validate output_filters references during config generation

### LINKED ISSUE #945: Validate output_filters references during config generation
While working on a similar ingress/egress processing flow in SGLang ([sgl-project/sglang#21154](https://github.com/sgl-project/sglang/pull/21154)), I noticed a small validation gap in Plano's filter-chain config.

  Plano already validates listener-level `input_filters` references during config generation, but `output_filters` references do not appear to get the same check.

  This means a typo in an output filter ID can make it past config generation and fail later at runtime, which is harder to debug.

  Example:

  ```yaml
  filters:
    - id: input_guard
      url: http://localhost:10500
      type: http

  listeners:
    - name: llm
      type: model
      port: 12000
      input_filters:
        - input_guard
      output_filters:
        - missing_output_guard

  I think output_filters should be validated against the same agents / filters ID set as input_filters.

  Business rationale: filter chains are part of Plano's guardrails story, so config mistakes should fail early with a clear error instead of surfacing later in the request path.

mukeshbaphna:
## Summary

Fixes #945.

This extends listener-level filter validation so `output_filters` references are checked during config generation, just like `input_filters`.

## Why

Filter chains are part of Plano's guardrails path. If an `output_filters` ID has a typo, config generation should fail early with a clear error instead of allowing the response path to continue without the intended filter.

This came up while working on a similar ingress/egress processing flow in SGLang: https://github.com/sgl-project/sglang/pull/21154

## Changes

- Validate listener-level `input_filters` and `output_filters` against defined `agents` / `filters`
- Add a config-generator regression test for an unknown `output_filters` ID

## Local Testing

Focused regression test:

```bash
cd cli
uv run pytest test/test_config_generator.py -k unknown_listener_output_filter
```

Full config-generator test file:

```bash
cd cli
uv run pytest test/test_config_generator.py
```

Result locally:

```text
21 passed
```

### DIFF
--- a/cli/planoai/config_generator.py
+++ b/cli/planoai/config_generator.py
@@ -562,15 +562,15 @@ def validate_and_render_schema():
                     "Please provide model_providers either under listeners or at root level, not both. Currently we don't support multiple listeners with model_providers"
                 )
 
-    # Validate input_filters IDs on listeners reference valid agent/filter IDs
+    # Validate listener-level filter IDs reference valid agent/filter IDs.
     for listener in listeners:
-        listener_input_filters = listener.get("input_filters", [])
-        for fc_id in listener_input_filters:
-            if fc_id not in agent_id_keys:
-                raise Exception(
-                    f"Listener '{listener.get('name', 'unknown')}' references input_filters id '{fc_id}' "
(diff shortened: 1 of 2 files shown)


mukeshbaphna:
Closing this draft in favor of #947, which uses a clean branch name.

PR #947, opened 2026-05-17
Validate model listener filter references before serving traffic

### LINKED ISSUE #945: Validate output_filters references during config generation

mukeshbaphna:
## Summary

Fixes #945.

This makes model-listener filter references fail closed in both config generation and Brightstaff startup. `output_filters` are now validated the same way as `input_filters`, and Brightstaff no longer silently drops missing filter IDs when resolving the runtime filter pipeline.

This PR does not change runtime output-filter failure behavior. It only makes invalid filter references fail fast before traffic is served. Startup/config validation and runtime stream execution need different handling modes: a missing filter ID is a config bug, while an output filter service failing mid-stream is still handled by the existing log-and-pass-through behavior.

## Why

Filter chains are part of Plano's guardrails path. If an `output_filters` ID has a typo, config generation or startup should fail early with a clear error instead of allowing the response path to continue without the intended filter.

This came up while working on a similar ingress/egress processing flow in SGLang: https://github.com/sgl-project/sglang/pull/21154

## Design Notes

- Python config generation catches the normal `plano_config.yaml` path before rendered config is produced.
- Brightstaff startup catches rendered configs or non-CLI startup paths that bypass Python config generation.
- Missing `input_filters` references already fail during config generation; this makes `output_filters` consistent with that behavior.
- Existing runtime behavior for configured output filters is unchanged: if an output filter errors while streaming, Plano logs the error and passes through the original chunk.

## Changes

- Validate listener-level `input_filters` and `output_filters` against defined `agents` / `filters` during config generation
- Preserve valid `output_filters` configs with a positive config-generator test
- Make Brightstaff startup return an error when rendered config references a missing filter ID
- Add focused Rust tests for valid, missing input filter, and missing output filter resolution

## Local Testing

```bash
cd cli
uv run pytest test/test_config_generator.py
```

```bash
cd crates
cargo test -p brightstaff resolve_filter_chain
```

Result locally:

```text
22 passed
3 passed
```

### DIFF
--- a/cli/planoai/config_generator.py
+++ b/cli/planoai/config_generator.py
@@ -562,15 +562,15 @@ def validate_and_render_schema():
                     "Please provide model_providers either under listeners or at root level, not both. Currently we don't support multiple listeners with model_providers"
                 )
 
-    # Validate input_filters IDs on listeners reference valid agent/filter IDs
+    # Validate listener-level filter IDs reference valid agent/filter IDs.
     for listener in listeners:
-        listener_input_filters = listener.get("input_filters", [])
-        for fc_id in listener_input_filters:
-            if fc_id not in agent_id_keys:
-                raise Exception(
-                    f"Listener '{listener.get('name', 'unknown')}' references input_filters id '{fc_id}' "
-                    f"which is not defined in agents or filters. Available ids: {', '.join(sorted(agent_id_keys))}"
-                )
+        for filter_field in ("input_filters", "output_filters"):
+            for fc_id in listener.get(filter_field, []):
+                if fc_id not in agent_id_keys:
+                    raise Exception(
+                        f"Listener '{listener.get('name', 'unknown')}' references {filter_field} id '{fc_id}' "
+                        f"which is not defined in agents or filters. Available ids: {', '.join(sorted(agent_id_keys))}"
+                    )
 
     # Validate model aliases if present
     if "model_aliases" in config_yaml:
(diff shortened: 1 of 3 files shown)


Spherrrical:
Thank you for the contribution! Will merge once CI is finished.

mukeshbaphna:
> Thank you for the contribution! Will merge once CI is finished.

Hi @Spherrrical - PR seems to be blocked by secret-backed e2e jobs. What's the recommendation?

Spherrrical (approved):

Spherrrical:
CI for fork PRs cannot access repository secrets (including after workflow approval), so e2e jobs will not get API keys on this PR.

I pushed the same branch to \`katanemo/plano\` and opened [#955](https://github.com/katanemo/plano/pull/955) to run full CI with secrets. Once #955 is green, this PR is safe to merge.

PR #957, opened 2026-05-24
test: add keyless model listener filter demo

mukeshbaphna:
## Summary

Adds a model-listener filter demo that exercises both input and output filters. The default path is fully local and requires no provider API keys or hosted model access. A second config lets developers run the same filter chain against the real OpenAI endpoint when they want provider-backed testing.

This follows up on #947 by proving the valid model-listener filter path end-to-end after invalid filter references now fail closed.

## Changes

- Add a fake OpenAI-compatible provider for deterministic local responses
- Add `config.openai.yaml` for optional real OpenAI testing
- Add an output redaction filter that removes sensitive markers from provider responses
- Wire the model-listener demo to use both `input_filters` and `output_filters`
- Make the compose setup select a config file via `PLANO_CONFIG_FILE`
- Extend the smoke test to cover allowed input, blocked input, and output redaction
- Add service-level pytest coverage for the local filter/provider services
- Update README with the developer value and local test command

## Why this helps developers

Developers can validate model-listener guardrails locally before using real providers. This catches filter wiring mistakes early, supports CI without secrets, and demonstrates request blocking plus response redaction without adding a passthrough agent layer.

## Local Testing

```bash
uv run --with pytest --with fastapi --with httpx --with pydantic python -m pytest demos/filter_chains/model_listener_filter/test_services.py -q
# 4 passed
```

```bash
cd cli && uv run pytest test/test_config_generator.py -q
# 22 passed
```

```bash
bash -n demos/filter_chains/model_listener_filter/test.sh
```

```bash
git diff --check
```

Also validated the updated demo config with `validate_and_render_schema`.
Also validated `config.openai.yaml` with `validate_and_render_schema`.

Note: local pre-push `npm run lint` currently fails on unrelated existing frontend lint issues in `packages/shared-styles` and `packages/ui`; this PR only changes `demos/filter_chains/model_listener_filter`.

### DIFF
--- a/demos/filter_chains/model_listener_filter/test.sh
+++ b/demos/filter_chains/model_listener_filter/test.sh
@@ -24,20 +24,37 @@ run_test() {
     local name="$1"
     local expected_code="$2"
     local body="$3"
+    local expected_body_contains="${4:-}"
+    local forbidden_body_contains="${5:-}"
 
     http_code=$(curl -s -o /tmp/plano_test_body -w "%{http_code}" \
         -X POST "$BASE_URL/chat/completions" \
         -H "Content-Type: application/json" \
         -d "$body")
 
-    if [ "$http_code" -eq "$expected_code" ]; then
-        echo "  PASS  $name (HTTP $http_code)"
-        PASS=$((PASS + 1))
-    else
+    if [ "$http_code" -ne "$expected_code" ]; then
         echo "  FAIL  $name — expected $expected_code, got $http_code"
         echo "        Body: $(cat /tmp/plano_test_body)"
         FAIL=$((FAIL + 1))
+        return
     fi
+
+    if [ -n "$expected_body_contains" ] && ! grep -Fq "$expected_body_contains" /tmp/plano_test_body; then
+        echo "  FAIL  $name — body did not contain '$expected_body_contains'"
+        echo "        Body: $(cat /tmp/plano_test_body)"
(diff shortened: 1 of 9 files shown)


mukeshbaphna:
Hi @Spherrrical @salmanap, this follow-up is ready for review when you have a chance. It builds on #947 by adding a local model-listener filter demo with both input and output filters, plus an optional OpenAI config for provider-backed testing.

demos/filter_chains/model_listener_filter/test.sh
```diff
@@ -48,19 +65,19 @@ run_test "Allowed request (math question)" 200 '{
...
   "model": "gpt-4o-mini",
   "messages": [{"role": "user", "content": "How to hack into a system"}],
   "stream": false
-}'
+}' "content_blocked"
 
-run_test "Allowed request (joke)" 200 '{
+run_test "Output filter redacts provider response" 200 '{
   "model": "gpt-4o-mini",
-  "messages": [{"role": "user", "content": "Tell me a joke"}],
```
mukeshbaphna, line 61: ... this is not to reduce coverage. I wanted to make sure the smoke test covered all three aspects: allow, block, and redact.

mukeshbaphna (commented):
added a comment to explain why I removed additional "safe request: joke, expects 200"

Spherrrical (approved):

PR #972, opened 2026-06-22
feat(tracing): provider-agnostic exporters with first-class PostHog support

Spherrrical:
## Summary

Adds a provider-agnostic telemetry export layer to brightstaff, driven by a new `tracing.exporters` config list, and ships PostHog as the first first-class integration. Point at your PostHog URL + project token and every LLM call is streamed as a `$ai_generation` event — as simple as LiteLLM's `posthog` callback.

```yaml
tracing:
  random_sampling: 100
  exporters:
    - type: posthog
      url: https://us.i.posthog.com
      api_key: $POSTHOG_API_KEY
      distinct_id_header: x-user-id   # optional; else anonymous
```

## Changes

- **Config**: new `tracing.exporters[]` (schema + `Exporter`/`PosthogExporter` in `common`), extensible to other backends via an internally-tagged enum.
- **PostHog exporter**: new `SpanExporter` (`posthog_exporter.rs`) that filters LLM spans, maps attributes to `$ai_*` properties (model, provider, latency, tokens, http status, trace ids), and POSTs a `/batch/` payload via reqwest. Failures are logged and dropped, never blocking requests.
- **Wiring**: `init_tracer` now composes multiple batch exporters (OTLP + each PostHog exporter); enable gate decoupled so PostHog works without `opentracing_grpc_endpoint`.
- **distinct_id**: resolved from a configurable request header into `AppState`, stamped on the span as `plano.distinct_id`; anonymous fallback (`\$process_person_profile=false`).
- **Span enrichment**: now records `llm.provider`. `capture_messages` opt-in sends the truncated user preview as `$ai_input`.
- **Docs/tests**: tracing guide + full reference config updated; Rust config + span-mapping tests and a CLI schema test added.

### DIFF
--- a/cli/test/test_config_generator.py
+++ b/cli/test/test_config_generator.py
@@ -388,6 +388,33 @@ model_providers:
 
 """,
     },
+    {
+        "id": "valid_tracing_posthog_exporter",
+        "expected_error": None,
+        "plano_config": """
+version: v0.4.0
+
+listeners:
+  - name: llm
+    type: model
+    port: 12000
+
+model_providers:
+  - model: openai/gpt-4o-mini
+    access_key: $OPENAI_API_KEY
+    default: true
+
+tracing:
+  random_sampling: 100
+  exporters:
+    - type: posthog
+      url: https://us.i.posthog.com
+      api_key: $POSTHOG_API_KEY
+      distinct_id_header: x-user-id
+      capture_messages: false
+
+""",
+    },
 ]
 
 
--- a/config/plano_config_schema.yaml
+++ b/config/plano_config_schema.yaml
@@ -447,6 +447,30 @@ properties:
             additionalProperties:
               type: string
         additionalProperties: false
+      exporters:
+        type: array
+        items:
+          oneOf:
+            - type: object
+              properties:
+                type:
+                  type: string
+                  const: posthog
+                url:
+                  type: string
+                api_key:
+                  type: string
+                distinct_id_header:
+                  type: string
+                enabled:
+                  type: boolean
+                capture_messages:
+                  type: boolean
+              additionalProperties: false
+              required:
+                - type
+                - url
+                - api_key
     additionalProperties: false
   mode:
     type: string
--- a/config/plano_config_schema.yaml
+++ b/config/plano_config_schema.yaml
@@ -462,8 +462,6 @@ properties:
                   type: string
                 distinct_id_header:
                   type: string
-                enabled:
-                  type: boolean
                 capture_messages:
                   type: boolean
               additionalProperties: false
--- a/crates/brightstaff/src/tracing/init.rs
+++ b/crates/brightstaff/src/tracing/init.rs
@@ -11,8 +11,8 @@ use tracing_subscriber::registry::LookupSpan;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::EnvFilter;
 
-use super::ServiceNameOverrideExporter;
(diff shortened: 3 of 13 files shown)


cli/test/test_config_generator.py
```diff
@@ -386,6 +386,33 @@ def test_validate_and_render_happy_path_agent_config(monkeypatch):
...
+    default: true
+
+tracing:
+  random_sampling: 100
+  exporters:
+    - type: posthog
+      url: https://us.i.posthog.com
+      api_key: $POSTHOG_API_KEY
+      distinct_id_header: x-user-id
+      capture_messages: false
```
adilhafeez, line 414: can we have multiple exporters of type posthog? If so whats the benefit of those exporters?

config/plano_config_schema.yaml
```diff
@@ -447,6 +447,30 @@ properties:
...
+                  type: string
+                  const: posthog
+                url:
+                  type: string
+                api_key:
+                  type: string
+                distinct_id_header:
+                  type: string
+                enabled:
+                  type: boolean
```
adilhafeez: whats the benefit of adding `enabled` when you can just remove that expoter?

config/plano_config_schema.yaml
```diff
@@ -447,6 +447,30 @@ properties:
...
         additionalProperties: false
+      exporters:
+        type: array
+        items:
+          oneOf:
+            - type: object
+              properties:
+                type:
+                  type: string
+                  const: posthog
```
adilhafeez, line 458: What about otel collector? Can I not use otel collector as exporter?

crates/brightstaff/src/tracing/init.rs
```diff
@@ -90,26 +90,56 @@ pub fn init_tracer(tracing_config: Option<&Tracing>) -> &'static SdkTracerProvid
...
+            // every span fans out to the OTLP collector and every exporter.
+            let mut builder = SdkTracerProvider::builder();
+
             // Create ServiceNameOverrideExporter to support per-span service names
             // This allows spans to have different service names (e.g., plano(orchestrator),
             // plano(filter), plano(llm)) by setting the "service.name.override" attribute
-            let exporter = ServiceNameOverrideExporter::new(endpoint);
+            if let Some(endpoint) = otel_endpoint.as_deref() {
+                builder = builder.with_batch_exporter(ServiceNameOverrideExporter::new(endpoint));
+            }
```
adilhafeez, line 128: what if otel collector is not defined? would posthog exporter still work?

cli/test/test_config_generator.py
Spherrrical: Yes, because posthog has EU + US deployments and also custom deployments. if users want to they can export to different posthog instances to comply for each instance (more customizability)

config/plano_config_schema.yaml
Spherrrical: Good point - will remove this!

config/plano_config_schema.yaml
Spherrrical: otel collector still goes via `tracing.opentracing_grpc_endpoint` exporters is specifically for backends that arent plain OTLP and need event translation (ex. posthog $ai_generation). u can send to collector and posthog at the same time

crates/brightstaff/src/tracing/init.rs
Spherrrical: yes it works without the collector

adilhafeez (approved):

PR #988, opened 2026-07-13
fix: GPT-5.6 SSE streams truncated through /v1/responses (chunk framing, raw identity passthrough, gzip window)

### LINKED ISSUE #980: [bug] SSE Streaming Response Truncated by UTF-8 Chunk Boundary — `JSON Parse error: Unterminated string`
# SSE Streaming Response Truncated by UTF-8 Chunk Boundary — `JSON Parse error: Unterminated string`

## Summary

When Plano forwards an Anthropic `/v1/messages` SSE streaming response that contains multi-byte UTF-8 characters (e.g. CJK text) in `input_json_delta` payloads, the stream is silently truncated mid-flight. The client receives a broken SSE stream and reports `API Error: JSON Parse error: Unterminated string`.

Root cause: `SseStreamIter::try_from` calls `std::str::from_utf8(bytes)?` on each raw chunk from `reqwest::bytes_stream()`. When a TCP/TLS chunk boundary lands inside a multi-byte UTF-8 character, `from_utf8` fails with `incomplete UTF-8 sequence`, the entire chunk is dropped, and downstream `content_block_delta` events never reach the client.

## Environment

- Plano version: 0.4.25
- Listener: `type: model` (LLM gateway), Anthropic Messages API passthrough
- Upstream: Anthropic-compatible `/v1/messages` SSE
- Client: Claude Code (Anthropic protocol)
- Payload: tool-use (`Write` tool) with Chinese content in `partial_json` deltas

## Error Observed by Client

```
API Error: JSON Parse error: Unterminated string
```

## Server-Side Log Evidence

From `docker logs plano` (gateway / `llm_gateway.wasm` warnings):

```
2026-06-26 10:21:46.472 [warning] wasm log http.config llm gateway: request id=2f8e963e-f5aa-9700-bf19-c46997425934: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 0

2026-06-26 10:21:52.623 [warning] wasm log http.config llm gateway: request id=185733b9-b66f-9918-89d0-47bb68975dba: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 409

2026-06-26 10:22:19.047 [warning] wasm log http.config llm gateway: request id=185733b9-b66f-9918-89d0-47bb68975dba: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 410

2026-06-26 10:22:19.161 [warning] wasm log http.config llm gateway: request id=185733b9-b66f-9918-89d0-47bb68975dba: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 417

2026-06-26 10:22:19.314 [warning] wasm log http.config llm gateway: request id=185733b9-b66f-9918-89d0-47bb68975dba: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 416

2026-06-26 10:24:30.250 [warning] wasm log http.config llm gateway: request id=38749dce-4323-9cb8-89e0-264b1e65b63f: sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index 405
```

The `index` values (0, 405, 409, 410, 416, 417) cluster around the ~4096-byte TCP/TLS chunk boundary — exactly where a 3-byte CJK character gets split.

## Upstream Response (Verified Complete)

The upstream response captured at the same request is **complete and well-formed** — it ends cleanly with `content_block_stop` → `message_delta` (`stop_reason: tool_use`) → `message_stop`. So the truncation happens inside Plano, not at the upstream.

Structure of the upstream SSE (8259 lines total, ~365 KB):

```
event: message_start
data:{"message":{"content":[],"id":"msg_vrtx_018iz9zC9Pus4RRec1rbsHXq","model":"anthropic.claude-opus-4-7","role":"assistant","type":"message","usage":{"cache_creation_input_tokens":9515,"cache_read_input_tokens":146439,"input_tokens":1,"output_tokens":3}},"type":"message_start"}

event: content_block_start
data:{"content_block":{"text":"","type":"text"},"index":0,"type":"content_block_start"}

event: content_block_delta
data:{"delta":{"text":"Now the","type":"text_delta"},"index":0,"type":"content_block_delta"}

... (text deltas) ...

event: content_block_stop
data:{"index":0,"type":"content_block_stop"}

event: content_block_start
data:{"content_block":{"id":"toolu_vrtx_01S3fFEMccXShNG6riR5BWqu","input":{},"name":"Write","type":"tool_use"},"index":1,"type":"content_block_start"}

event: content_block_delta
data:{"delta":{"partial_json":"","type":"input_json_delta"},"index":1,"type":"content_block_delta"}

event: content_block_delta
data:{"delta":{"partial_json":"{\"conten","type":"input_json_delta"},"index":1,"type":"content_block_delta"}

event: content_block_delta
data:{"delta":{"partial_json":"t\": \"#","type":"input_json_delta"},"index":1,"type":"content_block_delta"}

... (~8000 input_json_delta events with CJK partial_json content) ...

event: content_block_stop
data:{"index":1,"type":"content_block_stop"}

event: message_delta
data:{"delta":{"stop_details":null,"stop_reason":"tool_use","stop_sequence":null},"type":"message_delta","usage":{"cache_creation_input_tokens":9515,"cache_read_input_tokens":146439,"input_tokens":1,"output_tokens":11998,"output_tokens_details":{"thinking_tokens":0}}}

event: message_stop
data:{"type":"message_stop"}
```

The `partial_json` fields carry CJK characters (3-byte UTF-8) being streamed token-by-token as the model generates the tool-call input JSON. Example delta shape:

```
data:{"delta":{"partial_json":" <CJK text>","type":"input_json_delta"},"index":1,"type":"content_block_delta"}
```

### Failing code: `crates/hermesllm/src/apis/streaming_shapes/sse.rs:253-262`

```rust
impl TryFrom<&[u8]> for SseStreamIter<std::vec::IntoIter<String>> {
    type Error = Box<dyn std::error::Error + Send + Sync>;

    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
        let s = std::str::from_utf8(bytes)?;   // <-- fails on split multi-byte char
        let lines: Vec<String> = s.lines().map(|line| line.to_string()).collect();
        Ok(SseStreamIter::new(lines.into_iter()))
    }
}
```

`std::str::from_utf8` requires the entire input to be valid UTF-8. `reqwest::Response::bytes_stream()` yields chunks at arbitrary byte boundaries (typically ~4 KB). When a boundary falls inside a 3-byte CJK character, `from_utf8` returns `Utf8Error` with `incomplete UTF-8 sequence from index N`.

### Why the existing buffer doesn't help: `crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs:39-86`

```rust
pub fn process_chunk(&mut self, chunk: &[u8], ...) -> Result<Vec<SseEvent>, String> {
    let mut combined_data = std::mem::take(&mut self.incomplete_event_buffer);
    combined_data.extend_from_slice(chunk);

    let sse_iter = match SseStreamIter::try_from(combined_data.as_slice()) {
        Ok(iter) => iter,
        Err(e) => return Err(format!("Failed to create SSE iterator: {}", e)),
        //                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        //                UTF-8 error returned here — chunk dropped entirely
    };

    // ... the is_incomplete_json buffer below never runs for UTF-8 errors ...
    for sse_event in sse_iter {
        match SseEvent::try_from(...) {
            Ok(transformed) => { ... }
            Err(e) => {
                let is_incomplete_json = error_str.contains("eof while parsing")
                    || error_str.contains("unexpected end of json")
                    || error_str.contains("unexpected eof");
                if is_incomplete_json {
                    self.incomplete_event_buffer = sse_event.raw_line.as_bytes().to_vec();
                    break;
                } else {
                    continue;
                }
            }
        }
    }
}
```

The existing `is_incomplete_json` buffer (line 72) only handles incomplete *JSON*, and only after `SseStreamIter::try_from` has already succeeded. The UTF-8 failure happens **before** that, at line 50-53, so the buffer logic is never reached.

### Caller drops the chunk: `crates/llm_gateway/src/stream_context.rs:523-529`

```rust
Err(e) => {
    warn!(
        "request_id={}: sse chunk process error: {}",
        self.request_identifier(),
        e
    );
    return Err(Action::Continue);   // chunk silently dropped
}
```

Result: every chunk that straddles a UTF-8 character boundary is dropped. The client's SSE stream is missing `content_block_delta` events, and the reassembled `partial_json` JSON is truncated mid-string → `Unterminated string`.

## Why Anthropic Protocol + CJK Content Is the High-Frequency Trigger

- Anthropic Messages API streams tool-call inputs as `input_json_delta` events whose `partial_json` field carries raw model-generated text token-by-token.
- CJK characters are 3 bytes in UTF-8; emoji are 4 bytes. ASCII is 1 byte.
- A ~4 KB chunk boundary has a ~3× higher chance of splitting a CJK char than an ASCII char.
- OpenAI ChatCompletions streaming (mostly ASCII `delta.content`) rarely hits this because single-byte chars can't be split.

## Reproduction

1. Configure a `type: model` listener pointing at an Anthropic-compatible upstream.
2. Send a `/v1/messages` streaming request that triggers a tool-use response with substantial CJK content in the tool input (e.g. a `Write` tool writing a CJK markdown file, output ~10k+ tokens).
3. Observe:
   - Client receives `API Error: JSON Parse error: Unterminated string`
   - `docker logs plano` shows `sse chunk process error: Failed to create SSE iterator: incomplete UTF-8 sequence from index N`
   - Capturing the upstream response directly (bypassing Plano) shows a complete, well-formed SSE stream ending in `message_stop`

## Proposed Fix

Buffer the incomplete UTF-8 tail bytes in `SseChunkProcessor` instead of failing the whole chunk. Use `Utf8Error::valid_up_to()` to find the last complete character boundary.

```rust
pub fn process_chunk(&mut self, chunk: &[u8], ...) -> Result<Vec<SseEvent>, String> {
    let mut combined = std::mem::take(&mut self.incomplete_event_buffer);
    combined.extend_from_slice(chunk);

    // Find the last complete UTF-8 character boundary
    let safe_len = match std::str::from_utf8(&combined) {
        Ok(_) => combined.len(),
        Err(e) => {
            let valid = e.valid_up_to();
            if valid == 0 {
                // No valid UTF-8 prefix at all — buffer everything for next chunk
                self.incomplete_event_buffer = combined;
                return Ok(Vec::new());
            }
            valid
        }
    };

    // Process only the valid UTF-8 prefix; buffer the incomplete tail
    let (safe_bytes, tail) = combined.split_at(safe_len);
    self.incomplete_event_buffer = tail.to_vec();

    // safe_bytes is now guaranteed valid UTF-8
    let sse_iter = match SseStreamIter::try_from(safe_bytes) {
        Ok(iter) => iter,
        Err(e) => return Err(format!("Failed to create SSE iterator: {}", e)),
    };

    // ... rest of existing logic unchanged ...
}
```

Key change: incomplete UTF-8 tail bytes are buffered for the next chunk instead of failing the entire parse. The next chunk arrives, the multi-byte character is completed, and parsing proceeds normally.

### Verification after fix

```bash
cd crates && cargo build --release --target=wasm32-wasip1 -p llm_gateway
planoai down && planoai up config.yaml --docker
```

Reproduce the same request. `docker logs plano` should no longer show `incomplete UTF-8 sequence` warnings, and the client should no longer receive `Unterminated string`.

## Affected Code Locations

| Component | File:Line |
|---|---|
| UTF-8 validation failure point | `crates/hermesllm/src/apis/streaming_shapes/sse.rs:258` |
| Chunk processor entry (UTF-8 error escapes here) | `crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs:50-53` |
| Incomplete-JSON buffer (unreached for UTF-8 errors) | `crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs:66-80` |
| Caller drops chunk on error | `crates/llm_gateway/src/stream_context.rs:523-529` |

## Impact

- Any Anthropic Messages API streaming response with multi-byte UTF-8 content (CJK, emoji, accented Latin) can be truncated probabilistically based on TCP chunk boundaries.
- Severity scales with response size and non-ASCII density — long tool-use outputs with CJK content are most affected.
- The failure is silent from Plano's perspective (only a `warn` log) but fatal to the client stream.

Jiliac:
## Problem

`chatgpt/gpt-5.6-{sol,terra,luna}` (ChatGPT-subscription Codex backend) return HTTP 200 through Plano's `/v1/responses`, but the client only receives `response.created` + `response.in_progress` — no text, no `response.completed`. Fixes #980.

Live debugging with a from-source gateway traced this to **three stacked defects**, each independently able to truncate or corrupt streams:

### 1. SSE chunk-boundary loss in `SseChunkProcessor` (hermesllm)

`SseStreamIter` splits the buffer with `str::lines()` and parses each line independently. When a network chunk boundary lands inside a `data: `/`event: ` line *prefix* (e.g. `da` + `ta: {…}`), the partial line fails with a non-JSON error, which `SseStreamIter::next` silently skips — it never reaches the incomplete-JSON buffering. Both halves are unparseable, so the event is permanently lost with no log line. A boundary inside a multi-byte UTF-8 char could similarly kill the whole chunk via `str::from_utf8`.

**Fix:** byte-level SSE framing in `process_chunk` — only bytes up to the last `\n` are parsed; the trailing partial line is held back and prepended to the next chunk (with a completeness check for a final unterminated line like `data: [DONE]`, since nothing flushes the buffer at end-of-stream, and a 1 MiB cap so a broken upstream can't grow the buffer unboundedly).

### 2. Lossy parse-and-reconstruct on the identity Responses→Responses path

The "passthrough" path strictly deserialized every event into `ResponsesAPIStreamEvent` and re-serialized from the struct. Any unmodeled event type or field was dropped or mangled: via-Plano streams lost `reasoning.mode`/`reasoning.context`, `text.verbosity`, `tool_usage`, …, and a future unknown event type (e.g. `response.reasoning_text.delta`) would be dropped entirely. Additionally, `llm_gateway`'s per-event loop aborted the **whole remaining chunk** (`return Err(Action::Continue)`) when one event failed to parse.

**Fix:** on the identity Responses→Responses path, forward the original wire bytes verbatim (reconstructing the coupled `event: <type>\ndata: <json>\n\n` block from the raw payload — output is byte-identical to upstream). Parsing is still attempted for token counting; unknown events forward raw with `provider_stream_response = None`, and incomplete-JSON errors still propagate for cross-chunk recombination (shared `is_incomplete_json_error` helper, used by both call sites). `stream_context.rs` now logs and falls through on unparsed events instead of aborting the chunk. Cross-API arms and Anthropic/ChatCompletions identity behavior are untouched.

### 3. gzip decompressor `window_bits: 9` truncates streams (envoy config)

This turned out to be the dominant user-visible cause. Envoy's decompressor filter advertises `accept-encoding` upstream, so the backend returns gzip. The decompressor was configured with `window_bits: 9` (512-byte inflate window) while the paired compressor uses `window_bits: 10` and real upstreams use 15. Gzip headers don't carry the window size, so inflate doesn't fail upfront — it **silently stops emitting data** once a back-reference exceeds the window. Debug trace of a failing stream: `response data decompressed from 1372 bytes to 2638 bytes`, then `307 → 2`, then `0` for every remaining frame. This truncated *any* sufficiently large/redundant SSE stream traversing two listeners (gpt-5.4 included, when the client path involved compression).

**Fix:** decompressor `window_bits: 15` on all four gzip decompressor blocks (three were at 9; the `egress_traffic_llm` one had no `window_bits` and silently used Envoy's default of 12). Note the blast radius: this config applies to **all providers'** gzip-encoded responses, not just chatgpt — which is intended, since the truncation class affects any of them. Raising the inflate window is strictly safe (a larger window decodes anything a smaller one can; `max_inflate_ratio` zip-bomb protection is retained).

Also: added `gpt-5.6-{sol,terra,luna}` to the chatgpt models in `provider_models.yaml` (consumed via `include_str!` in `ProviderId`), and a pre-existing one-line clippy fix (`for_kv_map`) needed to keep the workspace `-D warnings` gate green.

## Verification

Unit tests (all inline, ids obfuscated):
- `test_gpt56_responses_identity_passthrough_full_stream` — replays a real captured GPT-5.6 stream (11 events incl. a reasoning output item) through the processor at **every 64-byte split offset**, asserting all event types emerge and output bytes == input bytes. Failed before the fix; passes now.
- Mid-prefix split, mid-JSON recombination, unknown-event tolerance, bounded-buffer, and serde relaxation tests (`OutputItem::Reasoning` without `summary` / with `content`/`encrypted_content`/`status`; omitted `logprobs`).
- Full gates: `cargo test --lib` (388 tests across brightstaff/common/hermesllm/llm_gateway/prompt_gateway), `cargo fmt --check`, `cargo clippy --locked --all-targets --all-features -- -D warnings`, wasm32-wasip1 release builds.

Live through a from-source gateway (`planoai build` + `up`):
- All three `chatgpt/gpt-5.6-*` tiers and `chatgpt/gpt-5.4` now stream the full ladder (`created … output_text.delta … completed`) with correct answer text.
- Previously-dropped fields (`reasoning.mode`, `text.verbosity`, `tool_usage`) are preserved byte-for-byte.

## Notes

- Wire-format change: identity Responses→Responses streams are now verbatim upstream bytes instead of Plano's re-serialized shape — strictly more spec-faithful; scoped to that identity pair only.
- Orthogonal to #958 (ChatGPT/5.5 path): this change deliberately doesn't touch the `ResponsesAPIStreamBuffer` cross-API path #958 works in.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/hermesllm/src/apis/streaming_shapes/sse.rs
+++ b/crates/hermesllm/src/apis/streaming_shapes/sse.rs
@@ -211,6 +211,20 @@ impl From<SseEvent> for Vec<u8> {
     }
 }
 
+/// Classify a transformation error as an "incomplete JSON" (EOF-while-parsing)
+/// error versus any other error (unknown event type, validation failure, etc.).
+///
+/// This is the single source of truth for that classification, shared by the
+/// SSE chunk processor (to decide whether to buffer a line for cross-chunk
+/// recombination) and the identity passthrough transform (to decide whether to
+/// propagate the error so the processor can retry with the next chunk).
+pub fn is_incomplete_json_error<E: std::fmt::Display + ?Sized>(err: &E) -> bool {
+    let msg = err.to_string().to_lowercase();
+    msg.contains("eof while parsing")
+        || msg.contains("unexpected end of json")
+        || msg.contains("unexpected eof")
+}
+
 #[derive(Debug)]
 pub struct SseParseError {
     pub message: String,
--- a/crates/hermesllm/src/apis/streaming_shapes/sse.rs
+++ b/crates/hermesllm/src/apis/streaming_shapes/sse.rs
@@ -211,18 +211,22 @@ impl From<SseEvent> for Vec<u8> {
     }
 }
 
-/// Classify a transformation error as an "incomplete JSON" (EOF-while-parsing)
-/// error versus any other error (unknown event type, validation failure, etc.).
-///
-/// This is the single source of truth for that classification, shared by the
-/// SSE chunk processor (to decide whether to buffer a line for cross-chunk
-/// recombination) and the identity passthrough transform (to decide whether to
-/// propagate the error so the processor can retry with the next chunk).
+/// Legacy display-text classifier retained for public API compatibility.
 pub fn is_incomplete_json_error<E: std::fmt::Display + ?Sized>(err: &E) -> bool {
     let msg = err.to_string().to_lowercase();
     msg.contains("eof while parsing")
         || msg.contains("unexpected end of json")
         || msg.contains("unexpected eof")
+}
+
+/// Classify a transformation error as incomplete JSON using serde's stable EOF
+/// category rather than the rendered error text. This is the single source of
+/// truth for internal streaming retry decisions.
+pub(crate) fn is_incomplete_json_error_from_error(
+    err: &(dyn Error + Send + Sync + 'static),
+) -> bool {
+    err.downcast_ref::<serde_json::Error>()
+        .is_some_and(serde_json::Error::is_eof)
 }
 
 #[derive(Debug)]
@@ -306,5 +310,45 @@ where
             }
         }
         None
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_is_incomplete_json_error_only_accepts_serde_eof() {
+        let incomplete_json = serde_json::from_str::<serde_json::Value>(r#"{"partial":"#)
+            .expect_err("truncated JSON must fail");
+        assert!(incomplete_json.is_eof());
+        let incomplete_json: Box<dyn Error + Send + Sync> = Box::new(incomplete_json);
+        assert!(is_incomplete_json_error_from_error(
+            incomplete_json.as_ref()
+        ));
+
+        let invalid_json = serde_json::from_str::<serde_json::Value>(r#"{"value":]}"#)
+            .expect_err("invalid JSON must fail");
+        assert!(!invalid_json.is_eof());
+        let invalid_json: Box<dyn Error + Send + Sync> = Box::new(invalid_json);
+        assert!(!is_incomplete_json_error_from_error(invalid_json.as_ref()));
+
+        let unrelated_eof =
+            std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "unexpected EOF");
+        let unrelated_eof: Box<dyn Error + Send + Sync> = Box::new(unrelated_eof);
+        assert!(!is_incomplete_json_error_from_error(unrelated_eof.as_ref()));
+    }
+
+    #[test]
+    fn test_is_incomplete_json_error_accepts_display_only_callers() {
+        struct DisplayOnly;
+
+        impl std::fmt::Display for DisplayOnly {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str("not an error")
+            }
+        }
+
+        assert!(!is_incomplete_json_error(&DisplayOnly));
     }
 }
--- a/crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
+++ b/crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
@@ -1,4 +1,6 @@
-use crate::apis::streaming_shapes::sse::{is_incomplete_json_error, SseEvent, SseStreamIter};
+use crate::apis::streaming_shapes::sse::{
+    is_incomplete_json_error_from_error, SseEvent, SseStreamIter,
+};
 use crate::clients::endpoints::{SupportedAPIsFromClient, SupportedUpstreamAPIs};
 
 /// Determine whether a trailing partial line (the bytes after the final `\n` in
@@ -11,10 +13,11 @@ use crate::clients::endpoints::{SupportedAPIsFromClient, SupportedUpstreamAPIs};
 /// processor at end-of-stream, so such a line must be parsed now rather than
 /// buffered forever.
 ///
-/// A line is "complete" when it parses as an SSE event and, for `data:` lines,
-/// its JSON payload is itself fully parseable (or is the `[DONE]` sentinel). A
-/// mid-prefix split (`da`), a mid-JSON split (`data: {"a":`), or a split inside
-/// a multi-byte UTF-8 character all fail this check and are held back.
+/// A trailing line is "complete" only when it is a `data:` line whose JSON
+/// payload is fully parseable (or is the `[DONE]` sentinel). Event-only lines
+/// are held back because a chunk can end midway through an `event:` name.
+/// A mid-prefix split (`da`), a mid-JSON split (`data: {"a":`), or a split
+/// inside a multi-byte UTF-8 character all fail this check and are held back.
 ///
 /// Soundness note: this relies on SSE data payloads being JSON *objects* (or
 /// `[DONE]`), where no strict prefix of a longer payload can itself fully
@@ -31,8 +34,9 @@ fn trailing_line_is_complete(bytes: &[u8]) -> bool {
             Some(data) => {
                 data.trim() == "[DONE]" || serde_json::from_str::<serde_json::Value>(data).is_ok()
             }
-            // event-only line (e.g. `event: foo`) with no data payload
-            None => true,
+            // Without a terminating newline, an event-only line may be a
+            // prefix of a longer event name, so retain it for the next chunk.
+            None => false,
         },
         Err(_) => false,
     }
@@ -155,7 +159,7 @@ impl SseChunkProcessor {
                     transformed_events.push(transformed);
                 }
                 Err(e) => {
-                    if is_incomplete_json_error(&e) {
+                    if is_incomplete_json_error_from_error(e.as_ref()) {
                         // Conservative fallback: with the framing above complete
                         // lines always carry complete JSON, so this should not
                         // trigger. If it ever does, buffer this line (plus the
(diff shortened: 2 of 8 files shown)


crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
```diff
@@ -1,6 +1,53 @@
...
+        // Split inside a multi-byte UTF-8 char -> definitely incomplete.
+        Err(_) => return false,
+    };
+    match s.parse::<SseEvent>() {
+        Ok(event) => match &event.data {
+            Some(data) => {
+                data.trim() == "[DONE]" || serde_json::from_str::<serde_json::Value>(data).is_ok()
+            }
+            // event-only line (e.g. `event: foo`) with no data payload
+            None => true,
```
Spherrrical: A chunk boundary landing inside an `event:` line (e.g. `event: respo`) parses as a valid event-only `SseEvent`, so `None => true` makes it "complete" and it gets consumed now with a truncated name; the continuation on the next chunk is then dropped as unparseable. This is safe today only because every coupled path re-derives the event name from `data.type` and suppresses the raw `event:` line — an implicit invariant, not an enforced one.

Could you either hold back a trailing `event:`-only line too (return `false` here when the line isn't newline-terminated), or add a comment documenting *why* truncating an `event:` line is safe, so a future non-coupled consumer doesn't silently regress?

crates/hermesllm/src/apis/streaming_shapes/sse.rs
```diff
@@ -211,6 +211,20 @@ impl From<SseEvent> for Vec<u8> {
...
 }
 
+/// Classify a transformation error as an "incomplete JSON" (EOF-while-parsing)
+/// error versus any other error (unknown event type, validation failure, etc.).
+///
+/// This is the single source of truth for that classification, shared by the
+/// SSE chunk processor (to decide whether to buffer a line for cross-chunk
+/// recombination) and the identity passthrough transform (to decide whether to
+/// propagate the error so the processor can retry with the next chunk).
+pub fn is_incomplete_json_error<E: std::fmt::Display + ?Sized>(err: &E) -> bool {
```
Spherrrical, line 215: Could you add a unit test pinning the actual `serde_json` EOF messages this must catch (so a wording change fails loudly), or classify on `serde_json::Error::is_eof()` where the concrete type is still available rather than the stringified `Box<dyn Error>`?

Spherrrical (commented):
Reviewed the code and ran the full local gate suite against this branch — `cargo test --lib` (176 tests across the 5 crates), `cargo fmt --check`, `cargo clippy --locked --all-targets --all-features -D warnings`, and the `wasm32-wasip1` release build all pass, including the 7 new regression tests. Nice work; the every-64-byte-offset replay test is a great guard.

A few non-blocking hardening/clarity asks below. Two things I couldn't cover locally and am trusting the from-source verification for: the live gzip `window_bits` behavior (needs a running Envoy + real upstream) and the e2e streaming path.

Also, on the `window_bits: 15` change — just flagging for ops awareness that this now affects gzip decoding for **all** providers, not only chatgpt. That's the correct scope given the truncation class, but it's not isolated to the reported model.

Nit: the full-stream test asserts `out.len() == baseline_len` + substring presence, but the PR description claims byte-identical output. Consider a strict `assert_eq!(out, expected_bytes)` if you want the test to actually enforce that claim.

crates/hermesllm/src/apis/streaming_shapes/sse_chunk_processor.rs
Jiliac: Fixed in 59e5428. Trailing event-only lines without a newline are now always retained, so a split such as `event: respo` cannot be consumed as a complete event. Added `test_mid_event_name_split_is_buffered_until_continuation`, which asserts both buffering after the first fragment and exact recombination after its continuation.

crates/hermesllm/src/apis/streaming_shapes/sse.rs
Jiliac: Fixed in 59e5428. Internal streaming retry decisions now downcast to `serde_json::Error` and use `is_eof()` rather than matching rendered error text. The public Display-based helper remains for API compatibility. Regression coverage confirms truncated JSON is retried, invalid JSON is not, and an unrelated `UnexpectedEof` is not misclassified.

Jiliac:
Addressed the remaining review items in 59e5428 and pushed the branch. The identity-stream fixture now asserts byte-for-byte equality for the single-chunk baseline and every 64-byte boundary phase. The gzip `window_bits: 15` all-provider scope remains documented in the PR description; it is intentionally global to gzip-encoded traffic through the affected listeners. Fresh verification passed: `cargo fmt --all -- --check`, workspace Clippy with warnings denied, `cargo test --lib`, and the `wasm32-wasip1` release build for both gateway plugins.

Spherrrical (approved):

Jiliac:
thanks @Spherrrical 🙏

PR #1009, opened 2026-08-05
Remove Arch-Function, prompt targets, and function calling

Spherrrical:
## Summary
- Remove the Arch-Function runtime path (`prompt_gateway` → `/function_calling` → hosted Arch-Function), including the brightstaff handler and CLI auto-inject of the `arch-function` provider.
- Drop `prompt_targets` from config/schema/common and simplify prompt_gateway to a passthrough filter.
- Delete related demos, docs, skills, and tests; keep the Envoy `plano` cluster for Plano-Orchestrator.

## Test plan
- [ ] `cargo test -p common --lib`
- [ ] `cargo test -p brightstaff --lib`
- [ ] `cargo check --target=wasm32-wasip1 -p prompt_gateway -p llm_gateway`
- [ ] Confirm generated configs no longer include `arch-function`
- [ ] Confirm `POST /function_calling` returns 404
- [ ] Smoke `planoai up` with a model-listener config (e.g. llm_gateway demo)


cli/test/source/failure.json
nehcgs, line 1: Let's keep `archfc.katanemo.dev` for now.

cli/test/source/success.json
nehcgs, line 1: Same, let's keep `archfc.katanemo.dev` for now.

crates/common/src/consts.rs
nehcgs, line 1: @Spherrrical For these headers, can we really remove them? We just want to kill Arch-Function, but not full function calling logic. The headers like `X_ARCH_STATE_HEADER` are general and applicable to any other models that support function calling. Double check here.

PR #1015, opened 2026-08-07
Opt-in: skip re-routing on non-user turns

Spherrrical:
## Problem

A coding agent answers one user turn with many LLM calls: the model replies with tool calls, the client runs them and posts results back for the next step. Every one of those steps re-entered quality routing, so a single turn could drift across models mid-generation.

That breaks the client — tool-call ids, reasoning state and the provider prompt cache all belong to the model that started the turn. Claude Code can't work around it: it never sends `X-Model-Affinity`, so the fix has to be server-side.

```text
before: user -> route -> model A -> tool result -> route -> model B  (breaks)
after:  user -> route -> model A -> tool result -> replay A          -> next user -> route again
```

This is loop-scoped reuse, not a permanent pin. A new user message always re-routes, so sequential / intent-based selection keeps working turn to turn.

**Off by default.** General Plano users keep per-request routing. IR (and anyone else who wants this) turns it on in config — no second Plano tree:

```yaml
routing:
  route_on_user_only: true
```

## Changes

**The gate** — `is_user_turn` checks only the tail of the normalized message list. A trailing `Role::User` runs the quality router; anything else (tool results, assistant steps, empty history) replays the prior decision. That is simpler and more robust than detecting a tool loop: sequential handoff happens when the user speaks again, including Anthropic packing new user text alongside `tool_result` (that normalizes to a trailing user message).

A trailing user message also has to carry something a user actually said. Claude Code injects `<system-reminder>` and `<user-prompt-submit-hook>` blocks as user-role text, and empty user content is reachable on every client API, so those envelopes are stripped and an empty remainder counts as harness output rather than an utterance. An uncaptioned attachment is real input and does route. ReAct-style `Observation:` text remains a known limitation — it is indistinguishable from user prose without string-matching a framework convention, so it re-routes.

**Opt-in** — both the proxy (`llm/mod.rs` Phase 3) and the decision endpoint (`routing_service.rs`) call `should_reuse_prior_decision(route_on_user_only, messages)`. When the flag is false, the router runs on every request. When true, a non-user tail replays the prior model and skips `router_chat_get_upstream_model`. It still calls `session_router::route`, so warmth, history and cost bookkeeping stay correct; the router sees the anchor as its own candidate and records no switch.

**Storage** — the implicit session key is derived for this path only when `route_on_user_only` is on *and* routing preferences can override the client model (plus the existing `prompt_caching` / `routing_budget` cases). Without a session key there is nowhere to store the decision on the default Claude Code path, and the request `model` field is the wrong fallback since clients never echo the routing override.

**Lane guard** — `SessionBinding` gains `requested_model`, recording what the client asked for. Reuse requires the lane to match, so Claude Code's `ANTHROPIC_SMALL_FAST_MODEL` side calls can't inherit the main loop's model even when they share a prompt prefix. This generalizes to any multi-lane client instead of hardcoding the `arch.claude.code.small.fast` alias. The lane is the alias-resolved model on both paths: the decision endpoint was using the raw request model, so with `model_aliases` configured it wrote bindings under the alias while the proxy read them under the target, and every step re-routed.

**Normalization** — the gate reads the normalized message list, so a client API that silently drops content changes routing. Bedrock Converse did: `get_messages` discarded `toolUse` and `toolResult` blocks, so a Bedrock tool step looked like a plain user message and re-routed every step, and dropped `image` blocks made an attachment-only turn normalize to empty text. Bedrock messages now split `toolResult` into `Role::Tool` messages, attach `toolUse` to the assistant as `tool_calls`, and carry images as `ContentPart::ImageUrl` — through the same helper the Anthropic path already used, so both providers normalize identically.

Reuse is declined — falling back to a normal route — on a cold session, a drifted prompt prefix, a different lane, or no session key (`X-Plano-Cache: off`). Every ambiguous case re-routes rather than risk a wrong pin.

Observable via the `plano.routing.skipped` span attribute (`not_user_turn`) and the `brightstaff_router_skips_total` counter.

**Also in this PR** — Responses compatibility, needed to test Codex locally:

- `tools[].type = "namespace"` is accepted, and null tool fields are omitted (OpenAI rejects `tools[].domains: null`)
- Responses `input` items dispatch on the `type` discriminator, so a coincidental `id` can no longer steal `function_call` / `function_call_output` into `ItemReference`
- `custom_tool_call` / `custom_tool_call_output` are first-class items, mapping to an assistant `tool_calls` entry and a `Role::Tool` message. Codex registers custom tools by default, so without this every Codex step normalized to a trailing user message and re-routed
- Reasoning items no longer emit `summary` / `content` when the client didn't send them: a reasoning item is opaque provider state the passthrough re-serializes verbatim, and adding keys can invalidate it (@nehcgs)
- A known tool-call type that fails its typed shape (a `function_call` with no `arguments`) is still dropped for Chat Completions, but now logged, since its `function_call_output` is left orphaned upstream (@nehcgs)
- Streaming Responses echoes the served model in its lifecycle events instead of `unknown`

## Tests

Driven through the real client-API parsers rather than hand-built message structs — the claim under test is that all three wire formats normalize to a trailing user / non-user role, so hand-building would assume the thing being verified.

- All three APIs skip routing on a tool / assistant tail *when the flag is on*
- Flag off: a tool tail does not skip (general Plano unchanged)
- Fresh user turns still route, including a packed Anthropic `tool_result` + new user text
- Empty and envelope-only user tails do not route, and an uncaptioned attachment does — checked on Chat Completions, Anthropic and Bedrock Converse
- A Codex `custom_tool_call_output` tail is a continuation, not a user turn
- A model alias resolves to its target before becoming the session lane
- Bedrock `toolUse` / `toolResult` survive normalization and round-trip back; an image round-trips as a data URL
- A malformed `function_call` degrades to a dropped item instead of failing the parse, and reasoning items don't invent absent arrays (@nehcgs)
- Replay works, including when no route matched on the first turn
- Haiku lane, cold miss, prefix drift and cache-off all fall back to routing
- Skipping the router does not skip session bookkeeping
- `routing.route_on_user_only` parses as a bool; absent / false is off, true is on

## Notes for review

- `docs/routing-api.md` claimed affinity "returns the cached model without re-running routing," which the code never did. Rewritten to describe user-turn replay, and `pinned` corrected to mean the session was warm.
- The Claude Code demo config sets `routing.route_on_user_only: true`. Stock configs do not.
- One shared `X-Model-Affinity` id is not enough for a client that makes side calls (a side chat, a summarizer, a subagent). A session key holds a single binding, so whichever lane ran last owns it: the lane and prefix guards keep the side call from being *pinned* to the main loop's model, but the loop's own pin is evicted and its next continuation re-routes. Implicit affinity is unaffected, since a side call's different system prompt derives its own key. Documented as one id per conversation, not one per client session. Keying bindings by `(session_id, requested_model)` so lanes coexist is deferred to a separate PR — it changes the cache-key shape and grows entries per session.
- Not covered: an end-to-end assertion that the orchestrator receives zero HTTP calls during a live Claude Code loop. That path has no test harness in the repo, so it needs a manual run against the demo config.

### DIFF
--- a/crates/hermesllm/src/transforms/request/from_bedrock.rs
+++ b/crates/hermesllm/src/transforms/request/from_bedrock.rs
@@ -1,8 +1,21 @@
 use crate::apis::amazon_bedrock::{
-    ContentBlock, ConversationRole, Message as BedrockMessage, ToolResultContentBlock,
+    ContentBlock, ConversationRole, ImageSource, Message as BedrockMessage, ToolResultContentBlock,
 };
-use crate::apis::openai::{FunctionCall, Message, MessageContent, Role, ToolCall};
+use crate::apis::openai::{
+    ContentPart, FunctionCall, ImageUrl, Message, MessageContent, Role, ToolCall,
+};
 use crate::clients::TransformError;
+use crate::transforms::request::from_anthropic::build_openai_content;
+
+/// Render a Bedrock image source as the data URL an OpenAI image part carries. The
+/// inverse of `parse_data_url` on the to-Bedrock side, so images round-trip.
+fn image_source_to_url(source: &ImageSource) -> String {
+    match source {
+        ImageSource::Base64 { media_type, data } => {
+            format!("data:{};base64,{}", media_type, data)
+        }
+    }
+}
 
 /// Flatten Bedrock tool result content blocks into the single text payload that an
 /// OpenAI `tool` message carries.
@@ -28,13 +41,22 @@ impl TryFrom<BedrockMessage> for Vec<Message> {
             ConversationRole::Assistant => Role::Assistant,
         };
 
-        let mut text_parts = Vec::new();
+        let mut content_parts = Vec::new();
         let mut tool_calls = Vec::new();
         let mut tool_results = Vec::new();
 
         for block in message.content {
             match block {
-                ContentBlock::Text { text } => text_parts.push(text),
+                ContentBlock::Text { text } => content_parts.push(ContentPart::Text {
+                    text,
+                    cache_control: None,
+                }),
+                ContentBlock::Image { image } => content_parts.push(ContentPart::ImageUrl {
+                    image_url: ImageUrl {
+                        url: image_source_to_url(&image.source),
+                        detail: None,
+                    },
+                }),
                 ContentBlock::ToolUse { tool_use } => {
                     tool_calls.push(ToolCall {
                         id: tool_use.tool_use_id,
@@ -51,9 +73,9 @@ impl TryFrom<BedrockMessage> for Vec<Message> {
                         flatten_tool_result_content(&tool_result.content),
                     ));
                 }
-                ContentBlock::Image { .. }
-                | ContentBlock::Document { .. }
-                | ContentBlock::GuardContent { .. } => continue,
+                // Documents have no OpenAI content-part equivalent, and guard content is
+                // Bedrock-side policy metadata rather than conversation content.
+                ContentBlock::Document { .. } | ContentBlock::GuardContent { .. } => continue,
             }
         }
 
@@ -70,11 +92,12 @@ impl TryFrom<BedrockMessage> for Vec<Message> {
             })
             .collect();
 
-        let text = text_parts.join("\n");
-        if !text.is_empty() || !tool_calls.is_empty() || result.is_empty() {
+        // Normalized through the same helper Anthropic uses, so a single text block
+        // collapses to plain text and an image survives as a content part on both.
+        if !content_parts.is_empty() || !tool_calls.is_empty() || result.is_empty() {
             result.push(Message {
                 role,
-                content: Some(MessageContent::Text(text)),
+                content: build_openai_content(content_parts, &tool_calls),
                 name: None,
                 tool_calls: if tool_calls.is_empty() {
                     None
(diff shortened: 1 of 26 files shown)


salmanap:
Is this a deterministic thing across v1/responses v1/chat/completions and v1/messages? We should be very certain here before making an implicit change to routing behavior because developers loose control here - although I will say it might be a very useful feature if done correctly.

docs/source/guides/llm_router.rst
nehcgs, line 1: `route_on_user_only`: When disabled (default), every LLM request is routed independently. When enabled, the router selects a model only at the start of a new user turn, and all subsequent requests within the same agentic loop are pinned to that model.

docs/routing-api.md
nehcgs, line 1: ## Per-request routing (default)
 
Applies when `route_on_user_only` is `false` (the default). Every request is routed independently, including loop continuations. A single user turn may therefore be served by multiple models. For example, a lightweight model handling routine tool-orchestration iterations while a stronger model is selected for a complex reasoning step or the final synthesis.
 
### When to use
 Use per-request routing for:
- **Best model per step.** Each step in a turn is served by the model best matched to its difficulty or specialty.
- **Cost efficiency.** Simple steps go to smaller models; only hard steps use expensive ones.
- **Escalation on failure.** A struggling model can be swapped out for the remainder of the turn.
- **Capacity flexibility.** Each request can be placed wherever capacity exists, with no pinning constraint.

## Turn-level routing
 
Applies when `route_on_user_only` is `true`. The router selects a model once per user turn and pins the remainder of the agentic loop to it.

### When to use
Use turn-level routing for:
- **Plan consistency.** One model carries its own reasoning and plan through the entire loop.
- **Cache locality.** Loop iterations reuse the shared prompt prefix cache; no switch-induced misses.
- **No state translation.** Avoids stripping or converting model-specific artifacts (e.g., signed thinking blocks).
- **Simpler parameter handling.** Engine-native parameters are resolved once per turn, not re-mapped per request.
 
### When routing occurs
 
A routing decision is made when the incoming request represents the **start of a new user turn**, defined as:
 
- The final message in the `messages` array has `role: "user"`, **and**
- Its content contains **no** `tool_result` blocks (Anthropic format) / the final message is not `role: "tool"` (OpenAI format).
In other words: the last message is genuine user text, not a tool result being fed back into an in-flight loop.
 
### When routing is skipped (sticky)
 
Routing is skipped and the previously selected model is reused when the request is a **continuation of an in-flight agentic loop**, detected by:
 
- The final message contains one or more `tool_result` blocks (or `role: "tool"`), **or**
- The final assistant message in the history contains `tool_use` blocks that have not yet been resolved.
In these cases the router pins the request to the model recorded for the current turn.

crates/hermesllm/src/transforms/request/from_bedrock.rs
nehcgs, line 1: It drops `Image` blocks, so an attachment-only normalizes to empty text. This is inconsistent with OpenAI/Anthropic. Should convert images to `ContentPart::ImageUrl` and add a test.

nehcgs (approved):

PR #1016, opened 2026-08-07
Add per-request max_switch_spend_pct header for routing budget

Spherrrical:
## Summary

- Adds **`x-routing-max-switch-spend-pct`** (alias **`max_switch_spend_pct`**) so callers can override `routing.routing_budget.max_switch_spend_pct` per request on the LLM proxy and routing-decision paths.
- Makes `max_switch_spend_pct` optional in config (defaults to **20%** when `routing_budget` is present); the header wins when set.
- If routing budget is not configured in YAML but the header is present, a one-off effective budget is synthesized with default replenish/cache-read settings (still needs a cost metrics source for gating to work).

## Behavior

| Header | Config `routing_budget` | Result |
|--------|-------------------------|--------|
| absent | absent | Budget off |
| absent | present | Config cap (default 20% if omitted) |
| present | present | Header cap |
| present | absent | Header cap with default side settings |

Invalid header values are logged and ignored (config/default applies).

## Test plan

- [ ] `cargo test --lib -p common -p brightstaff` (routing budget + header parsing tests)
- [ ] Manual: `routing_budget: {}` in config + `x-routing-max-switch-spend-pct: 5` on a session with affinity; verify switch veto/allow vs baseline

[Slack Thread](https://digitalocean.slack.com/archives/C0BEDCC09NG/p1786140857862889?thread_ts=1786140857.862889&cid=C0BEDCC09NG)

<div><a href="https://cursor.com/agents/bc-bec73235-84e7-579e-9590-5d0cf73c71c0?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-bec73235-84e7-579e-9590-5d0cf73c71c0&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/crates/common/src/configuration.rs
+++ b/crates/common/src/configuration.rs
@@ -305,7 +305,7 @@ pub struct RoutingBudget {
     /// `0` means "never pay to switch" (only outright-cheaper switches are ever
     /// allowed); larger values buy more quality-driven switches. Typical range 10–30.
     /// When omitted, defaults to [`DEFAULT_MAX_SWITCH_SPEND_PCT`]; per-request overrides
-    /// use [`crate::consts::PLANO_MAX_SWITCH_SPEND_PCT_HEADER`].
+    /// use [`crate::consts::ROUTING_MAX_SWITCH_SPEND_PCT_HEADER`].
     #[serde(default)]
     pub max_switch_spend_pct: Option<f64>,
     /// Reset the running baseline/spend totals when a session goes cold and re-binds
@@ -929,7 +929,8 @@ mod test {
 
     use super::{
         EffectivePromptCaching, EffectiveRoutingBudget, IntoModels, LlmProvider, LlmProviderType,
-        PromptCaching, RoutingBudget, DEFAULT_CACHE_READ_DISCOUNT, DEFAULT_MIN_PREFIX_TOKENS,
+        PromptCaching, RoutingBudget, DEFAULT_CACHE_READ_DISCOUNT, DEFAULT_MAX_SWITCH_SPEND_PCT,
+        DEFAULT_MIN_PREFIX_TOKENS,
     };
     use crate::api::open_ai::ToolType;
 
--- a/crates/common/src/configuration.rs
+++ b/crates/common/src/configuration.rs
@@ -346,6 +346,23 @@ pub struct EffectiveRoutingBudget {
 pub const DEFAULT_CACHE_READ_DISCOUNT: f64 = 0.1;
 /// Default overhead cap when `max_switch_spend_pct` is omitted from config (20%).
 pub const DEFAULT_MAX_SWITCH_SPEND_PCT: f64 = 20.0;
+/// Inclusive bounds for `max_switch_spend_pct` (percent of never-switch baseline).
+pub const MAX_SWITCH_SPEND_PCT_MIN: f64 = 0.0;
+pub const MAX_SWITCH_SPEND_PCT_MAX: f64 = 100.0;
+
+/// Returns `Ok(())` when `pct` is a finite percent in [`MAX_SWITCH_SPEND_PCT_MIN`],
+/// [`MAX_SWITCH_SPEND_PCT_MAX`].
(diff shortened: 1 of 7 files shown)


crates/common/src/consts.rs
```diff
@@ -29,6 +29,11 @@ pub const PLANO_CACHE_HEADER: &str = "x-plano-cache";
 /// Hash of the stable prompt prefix, forwarded upstream so self-hosted multi-replica
 /// backends can do KV-aware (consistent-hash) replica routing at the LB/Envoy layer.
 pub const PLANO_PREFIX_HASH_HEADER: &str = "x-plano-prefix-hash";
+/// Per-request override for `routing.routing_budget.max_switch_spend_pct` (whole number,
+/// e.g. `20` = 20%). When absent, the configured default applies.
+pub const ROUTING_MAX_SWITCH_SPEND_PCT_HEADER: &str = "x-routing-max-switch-spend-pct";
```
adilhafeez: name it with plano prefix line x-plano

crates/common/src/consts.rs
```diff
@@ -29,6 +29,11 @@ pub const PLANO_CACHE_HEADER: &str = "x-plano-cache";
...
+/// Alias (field name forwarded by some proxies).
+pub const MAX_SWITCH_SPEND_PCT_HEADER_ALIAS: &str = "max_switch_spend_pct";
```
adilhafeez: why do we need alias for same field? let's remove it don't want any confusion

crates/common/src/configuration.rs
```diff
@@ -341,15 +344,20 @@ pub struct EffectiveRoutingBudget {
 }
 
 pub const DEFAULT_CACHE_READ_DISCOUNT: f64 = 0.1;
+/// Default overhead cap when `max_switch_spend_pct` is omitted from config (20%).
+pub const DEFAULT_MAX_SWITCH_SPEND_PCT: f64 = 20.0;
```
adilhafeez: why 20%? we should require user to specify the value let's not assume and fail here if this value is missing

crates/common/src/configuration.rs
```diff
@@ -341,15 +344,20 @@ pub struct EffectiveRoutingBudget {
...
 
 impl RoutingBudget {
     /// Resolve to effective settings, validating the overhead cap and cache-read
     /// discount.
     pub fn resolve(&self) -> Result<EffectiveRoutingBudget, String> {
-        if !self.max_switch_spend_pct.is_finite() || self.max_switch_spend_pct < 0.0 {
+        let max_switch_spend_pct = self
+            .max_switch_spend_pct
+            .unwrap_or(DEFAULT_MAX_SWITCH_SPEND_PCT);
+        if !max_switch_spend_pct.is_finite() || max_switch_spend_pct < 0.0 {
```
adilhafeez: should you not check it to be between 0.00 and 100.00?

adilhafeez (changes requested):

crates/common/src/consts.rs
Spherrrical: It was suggested by Khannan to set it as `x-routing...`

> "can we avoid the "plano" prefix since this is also directly exposed in Inference Router?"

In the Slack channel.

crates/common/src/configuration.rs
Spherrrical: 20% is a default value we've added in our demos and documentation as a way of showing. We will let the user decide to set whatever they want when it's sent upstream via the header.

adilhafeez (approved):

PR #1018, opened 2026-08-13
fix: prevent divide-by-zero panic in streaming token throughput metrics

### LINKED ISSUE #999: [BUG] Integer divide-by-zero panic in handle_end_of_request_metrics_and_traces
# Integer divide-by-zero panic in `handle_end_of_request_metrics_and_traces`

## Summary

A guaranteed integer divide-by-zero panic in `crates/llm_gateway/src/stream_context.rs:387` crashes the entire WASM VM when a streaming response is both fast and produces many output tokens. The panic manifests as a `RuntimeError` trap in the Envoy WASM sandbox.

## Environment

- **File**: `crates/llm_gateway/src/stream_context.rs`
- **Function**: `handle_end_of_request_metrics_and_traces` (line 360)
- **Version**: `0.4.29` (and all prior versions containing this code path)

## Root Cause

In `handle_end_of_request_metrics_and_traces`, lines 376–391:

```rust
if self.response_tokens > 0 {
    // Compute the time per output token
    let tpot = duration_ms as u64 / self.response_tokens as u64;  // line 378

    // Record the time per output token
    self.metrics.time_per_output_token.record(tpot);

    info!(
        "request_id={}: token throughput, time_per_token={}ms tokens_per_second={}",
        self.request_identifier(),
        tpot,
        1000 / tpot                                                // line 387 — PANIC
    );
    // Record the tokens per second
    self.metrics.tokens_per_second.record(1000 / tpot);           // line 390 — PANIC
}
```

**Step-by-step breakdown:**

1. **Line 378**: `tpot = duration_ms / response_tokens`. Both are integers (`u64`). When the response is fast and produces many tokens — e.g., `duration_ms = 790`, `response_tokens = 982` — integer division yields `790 / 982 = 0`.

2. **Line 387**: `1000 / tpot` → `1000 / 0`. Rust integer division by zero is a **guaranteed panic**, not UB and not `NaN`/`Inf` (unlike floating-point).

3. **WASM trap**: The panic triggers `abort()` → `unreachable` WASM instruction → the entire WASM VM is trapped and killed by the Envoy host. The request is lost, and subsequent requests routed to the same worker will also fail until the VM is restarted.

## Trigger conditions

- **Streaming response** (or non-streaming) where `response_tokens > 0` and `duration_ms < response_tokens` — i.e., any response faster than ~1 token per millisecond.
- This is hit regularly in production with high-throughput/low-latency streaming responses, where token count is estimated via `content.len() / 4` and easily exceeds the millisecond latency.

Examples from logs:

```
latency=915ms, tokens=1283  →  tpot = 915 / 1283 = 0  →  1000 / 0 = panic
latency=790ms, tokens=982   →  tpot = 790 / 982  = 0  →  1000 / 0 = panic
```

The panic backtrace in WASM points to `stream_context.rs:387:25`, which is column 25 — exactly the `1000 / tpot` expression.

## Impact

- **Severity**: Critical — crashes the WASM sandbox, killing in-flight requests and all queued requests for that worker.
- **Scope**: Any streaming (or non-streaming) response where `response_tokens * 1ms > duration_ms`. In practice, many normal streaming completions trigger this.

## Fix

Two changes:

1. **Guard against zero `tpot`** — add `duration_ms > 0` to the condition so `tpot = 0` doesn't cascade.
2. **Compute `tokens_per_second` directly** — use `response_tokens * 1000 / duration_ms` instead of `1000 / tpot`, so the metric is accurate even when `tpot` rounds to zero via integer division.

```rust
if self.response_tokens > 0 && duration_ms > 0 {
    let tpot = duration_ms as u64 / self.response_tokens as u64;

    self.metrics.time_per_output_token.record(tpot);

    let tokens_per_second =
        self.response_tokens as u64 * 1000 / duration_ms as u64;

    info!(
        "request_id={}: token throughput, time_per_token={}ms tokens_per_second={}",
        self.request_identifier(),
        tpot,
        tokens_per_second
    );
    self.metrics.tokens_per_second.record(tokens_per_second);
}
```

`tpot` is still recorded (it can legitimately be `0` for sub-millisecond-per-token streams). The `tokens_per_second` computation avoids the intermediate zero and produces a correct, non-zero value for all valid inputs.

SomSamantray:
Fast, short streaming completions (more output tokens than elapsed milliseconds) crashed the `llm_gateway` WASM worker with a divide-by-zero panic while recording token-throughput metrics, failing that request and any others sharing the worker. The throughput calculation now always records `time_per_output_token` safely, and only skips `tokens_per_second` (with a debug-level log noting the skip) for the rare case where elapsed time per token rounds to zero, instead of dividing by zero.

Fixes #999.

Verified with 5 new unit tests covering the exact regression scenario from the issue plus boundary cases (`cargo test --lib -p llm_gateway`), along with `cargo fmt --check`, `cargo clippy -D warnings`, and the `wasm32-wasip1` release build.

### DIFF
--- a/crates/llm_gateway/src/stream_context.rs
+++ b/crates/llm_gateway/src/stream_context.rs
@@ -374,20 +374,30 @@ impl StreamContext {
                 self.metrics.request_latency.record(duration_ms as u64);
 
                 if self.response_tokens > 0 {
-                    // Compute the time per output token
-                    let tpot = duration_ms as u64 / self.response_tokens as u64;
+                    let (tpot, tokens_per_second) =
+                        compute_token_throughput(duration_ms as u64, self.response_tokens as u64);
 
-                    // Record the time per output token
                     self.metrics.time_per_output_token.record(tpot);
 
-                    info!(
-                        "request_id={}: token throughput, time_per_token={}ms tokens_per_second={}",
-                        self.request_identifier(),
-                        tpot,
-                        1000 / tpot
-                    );
-                    // Record the tokens per second
-                    self.metrics.tokens_per_second.record(1000 / tpot);
+                    match tokens_per_second {
+                        Some(tokens_per_second) => {
+                            info!(
+                                "request_id={}: token throughput, time_per_token={}ms tokens_per_second={}",
+                                self.request_identifier(),
+                                tpot,
+                                tokens_per_second
+                            );
+                            self.metrics.tokens_per_second.record(tokens_per_second);
+                        }
+                        None => {
+                            // See compute_token_throughput: tpot==0 means we skip
+                            // tokens_per_second rather than divide by zero.
+                            debug!(
+                                "request_id={}: token throughput, time_per_token=0ms, tokens_per_second skipped (tpot==0)",
+                                self.request_identifier()
+                            );
+                        }
+                    }
                 }
             }
             Err(e) => {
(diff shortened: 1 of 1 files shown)


Spherrrical (approved):
Thank you for the contribution!

